PR #1, opened 2018-01-19
Initial serving of tile data over WCS

jpolchlo:
This PR builds out the WCS server's ability to serve tile data.  This is only a first pass attempt which serves data in WGS84, regardless of the layer projection.  Layers are assumed to consist of spatially-keyed singleband tiles.

That said, we have tile display in QGIS:
qgis-nlcd

Signed-off-by: jpolchlo <jpolchlopek@azavea.com>

### DIFF
--- a/src/server/src/main/scala/geotrellis/server/wcs/WcsRoute.scala
+++ b/src/server/src/main/scala/geotrellis/server/wcs/WcsRoute.scala
@@ -2,6 +2,8 @@ package geotrellis.server.wcs
 
 import geotrellis.server.wcs.params._
 import geotrellis.server.wcs.ops._
+import geotrellis.spark._
+import geotrellis.spark.io._
 
 import akka.http.scaladsl.marshallers.xml.ScalaXmlSupport._
 import akka.http.scaladsl.server._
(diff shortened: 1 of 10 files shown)


src/server/src/main/scala/geotrellis/server/wcs/ops/GetCoverage.scala
```diff
@@ -1,7 +1,62 @@
...
+import geotrellis.spark.stitch._
+
+import com.typesafe.config.ConfigFactory
+
+import scala.util.Try
 
 object GetCoverage {
-  def build(params: GetCoverageWCSParams): Array[Byte] = ???
+  def build(catalog: WcsRoute.MetadataCatalog, params: GetCoverageWCSParams): Array[Byte] = {
+    def valueReader = Try(ConfigFactory.load().getString("server.catalog")).toOption match {
```
echeipesh: You'd want to use the [`S3CollectionLayerReader`](https://github.com/locationtech/geotrellis/blob/master/s3/src/main/scala/geotrellis/spark/io/s3/S3CollectionLayerReader.scala) to parallelize over fetching of the tiles.

echeipesh: You'll probably need to write an SPI interface for `CollectionLayerReader`

src/server/src/main/scala/geotrellis/server/wcs/WcsRoute.scala
```diff
@@ -2,43 +2,78 @@ package geotrellis.server.wcs
...
-                }
-            }
+    get { ctx =>
+      extractUri { uri =>
+        println(s"Request received: $uri")
+        parameterMultiMap { params =>
+          WCSParams(params) match {
+
+            case Invalid(errors) =>
+              complete {
```
echeipesh: You'll want to `complete { Future { ... } }` in some way here.

jpolchlo:
The summary of @echeipesh's review is "works fine, can improve, but no reason it should not be merged right now, as it is usable".  The main complaint about performance is to better use concurrency.  Largely by using CollectionLayerReaders.  This is captured in #4, and will be attended to at the first opportunity.  Merging now, will intend to address the points of weakness in future updates.

PR #8, opened 2018-06-14
Support COGs

moradology:
This PR should go in after the http4s refactor and is currently heavily WIP.


moradology:
This PR Adds support for a *very* limited subset of the COG spec. We'll probably need to think things through more carefully before extending support to non-RGB cogs (perhaps leaning on MAML)

moradology:
Pulling the trigger on this so that development can push forward

PR #9, opened 2018-06-19
[WIP] Caching

### LINKED ISSUE #11: Caching interface (with sugar)
We need to provide a simple way to specify a caching strategy at runtime. This likely means specifying such a mechanism at the command line or via configuration (env vars or application.conf).

We'll need to support at least 4 cases to begin with:
1. NO CACHE: useful for debugging and thinking about performance
2. IN MEMORY: useful for quick demonstration and small services
3. MEMCACHED: useful for slightly more robust, single server cases
4. HYBRID: useful for multiple tile servers which share a memcached instance

moradology:
resolves #11


moradology:
This PR is stale and got a bit too clever to be taken seriously; should rewrite with something less flashy

PR #19, opened 2018-06-27
Add maml support [WIP]

moradology:
This is a WIP PR aimed at fully supporting MAML-based endpoints; both dynamically allocated via some store which backs them or statically defined during the server instantiation. Special attention is being paid to reusability, so any concerns raised related to reuse would be highly appreciated


moradology:
Closing in favor of new PR (which is based on upstream changes to MAML)

PR #21, opened 2018-07-12
Example maml services

### LINKED ISSUE #10: Interface for layer/tile-fetch-function reference
Because the GT server is supposed to be pluggable and generic, we can't strictly mandate the types of objects/information users will ultimately design their programs around. This means that we can't assume, for instance, that `UUID`s will always uniquely refer to layers. What people will provide is some means of referring to tiles and some proof that they can produce a tile on the basis of said reference; what we will provide is a means of using the reference, itself, as though it were tile data.

The existing MAML solution in Raster Foundry is actually quite close to this behavior already and such a structure happens to be useful for optimizing MAML resolution. In brief, we:
1. collect all sources from a given AST
2. only keep distinct elements (the set of sources)
3. resolve a tile/region corresponding to each source node in the set
4. rewrite the AST with tile literals in preparation for evaluation

Likely output for this transformation process:
 - TMS: `(Int, Int, Int) => Tile`
 - Batch/WCS/TIFF: `Extent => Raster`

Cf. https://github.com/raster-foundry/raster-foundry/blob/develop/app-backend/tile/src/main/scala/tool/TileResolver.scala

moradology:
This PR should expose an example `MAML` service which is semi-configurable as a means of COG support. The cog resolution logic requires some further efforts to avoid timeouts (caching headers necessary?) before the PR is complete

Closes #10


core/src/main/scala/geotrellis/server/core/maml/CogDefinition.scala
```diff
@@ -0,0 +1,51 @@
...
+import geotrellis.raster._
+import geotrellis.vector.Extent
+
+import java.net.URI
+
+
+case class CogNode(uri: URI, band: Int, celltype: Option[CellType])
+
+object CogNode {
+   private val intNdTile = IntConstantTile(NODATA, 256, 256)
```
pomadchin: wondering is it really good to return nodata tiles? leaflet recognizes `404` err code. (`Option[Tile]` for instance)

http4s/src/main/scala/geotrellis/server/http4s/maml/ExampleMamlService.scala
```diff
@@ -0,0 +1,90 @@
+package geotrellis.server.http4s.maml
```
pomadchin: looking towards keeping this example into the branch, mb it makes sense to move it into `examples` package?

http4s/src/test/scala/geotrellis/server/http4s/maml/GeneratedTileNode.scala
```diff
@@ -0,0 +1,51 @@
...
+import cats.syntax.all._
+import geotrellis.raster._
+import geotrellis.vector.Extent
+
+import java.net.URI
+
+
+case class GeneratedTileNode(mapper: Int => Int, size: Int = 256)
+
+//object GeneratedTileNode {
```
pomadchin: Hm there are lots of commented out code lines, do you want to include them into master branch? I know it's in a very R&D phase but still.

core/src/main/scala/geotrellis/server/core/maml/package.scala
```diff
@@ -0,0 +1,24 @@
...
+import cats._
+import cats.effect.IO
+import io.circe._
+import com.googlecode.concurrentlinkedhashmap.ConcurrentLinkedHashMap
+
+import java.util.UUID
+
+
+package object maml {
+  type HashMapMamlStore = ConcurrentLinkedHashMap[UUID, Expression]
```
pomadchin: Do we really need there `ConcurrentLinkedHashMap`?

moradology: Perhaps not, but it makes demonstration very simple

http4s/src/test/scala/geotrellis/server/http4s/maml/GeneratedTileNode.scala
moradology: Yeah, this is WIP - not going to merge in comments that aren't meant to be there

moradology:
Merging to year 0

PR #22, opened 2018-07-12
Add authorization to services

### LINKED ISSUE #16: Auth
RFML has JWT-specific auth requirements. One thing we should do early on is figure out how this will impact decisions elsewhere. To do this, we should produce a simple (maybe probabilistic?) authorization mechanism and play around in the fallout

jisantuc:
Overview
-----

This PR adds simple authorization with JWT or a few other strategies to services. It includes debug
output to verify that authorization works. In practice, we'd expand on configurability in the auth keys
to let users specify their own flows for obtaining an authorization workflow.

Testing
-----

- create a random signing key
- `export SIGNING_KEY=thatkey`
- run the server
- copy the jwt it offers you
- make an authed request to `/ping`
- it should succeed
- make an unauthed request to `/ping`
- it should fail

Closes #20 

Closes #16


moradology:
LGTM. As things move forward I anticipate some organizational efforts being useful/necessary with auth, but I'm not sure exactly what that should look like yet

PR #27, opened 2018-08-08
Use relative paths where possible in client code

moradology:
Client code won't work in deployment if we don't use relative paths (we have no idea who the host will be). This PR addresses that shortcoming


moradology:
trivial; merging upon travis success

PR #31, opened 2018-10-01
Upgrade cats -> 1.3.1 and cats-effect -> 1.0 and fix fallout

### LINKED ISSUE #30: Upgrade cats core and cats effect
`cats` and `cats-effect` both need to be pretty current to use `meow-mtl` (1.3.1 and 1.0, respectively), which we'd like to use in Raster Foundry land for error-handling. There are some major changes in those libraries from the versions in `geotrellis-server` that prevent upgrade without `geotrellis-server` also upgrading.

jisantuc:
Overview
-----

This PR upgrades cats and cats-effect to current versions. Main changes are:

- `parFoo` methods were replaced with `Foo`, since I'm not sure right now how to provide evidence for `Parallel[IO, F]`
- `implicit t: Timer[IO]` -> `implicit contextShift: ContextShift[IO]`, since the latter is what's required for evidence in `cats-effect` 1.0

Testing
-----

- compilation is fine
- run some examples

Closes #30

### DIFF
--- a/core/src/main/scala/geotrellis/server/core/maml/MamlExtent.scala
+++ b/core/src/main/scala/geotrellis/server/core/maml/MamlExtent.scala
@@ -29,7 +29,7 @@ object MamlExtent extends LazyLogging {
   )(
     implicit reify: MamlExtentReification[Param],
              enc: Encoder[Param],
-             t: Timer[IO]
+             contextShift: ContextShift[IO]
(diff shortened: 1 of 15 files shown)


moradology:
Looks like `ContextShift` was enough to get parallelism working. I took the liberty of reverting the changes which impacted parallel execution.

moradology:
The parallelism seems fine with your `ContextShift` change, so I've gone back and re-parallelized

PR #52, opened 2018-11-12
Enable access to avro-encoded GT layers

jpolchlo:
This PR provides the facilities to source imagery from GT avro layers.


echeipesh:
@moradology Is this safe to merge now ?

PR #54, opened 2018-11-20
Multiband in GT server WIP

moradology:
This brings us pretty close to supporting Multiband rasters (and therefore retrieval via gdal rather than through the JVM). Further testing of all examples will likely be required and MAML changes might be necessary as well


moradology:
@pomadchin  This WIP is being developed against a locally published https://github.com/geotrellis/maml/pull/70

moradology:
Superceded by #56

PR #56, opened 2018-11-26
GeoTrellis-Contrib GDAL integration

pomadchin:
It's a first attempt to use GDAL there to perform server operations. In this PR there is a new `GDALNode` object implemented. 

TODO: we need to move towards `geotrellis-contrib` project usage, it means that we can abstract in some cases over `RasterSources` and it would mean that we can support both `GeoTrellis GeoTiff` and `GDAL` `RasterSources` backends.


pomadchin:
Btw this already should work, at least NDVI example. NOTE: percent encode jsons before feeding them to ndvi routes.

pomadchin:
@moradology that's right! will fix it

core/src/main/scala/geotrellis/server/vlm/geotiff/CogNode.scala
```diff
@@ -47,14 +47,12 @@ object CogNode extends RasterSourceUtils {
...
       def fetch(xCoord: Int, yCoord: Int) =
-        CogUtils.fetch(self.uri.toString, z, xCoord, yCoord)
-        .map(_.tile)
-        .map(_.band(self.band))
-      (fetch(x - 1, y + 1), fetch(x, y + 1), fetch(x + 1, y + 1),
-       fetch(x - 1, y),     fetch(x, y),     fetch(x + 1, y),
-       fetch(x - 1, y - 1), fetch(x, y - 1), fetch(x + 1, y - 1)).parMapN { (tl, tm, tr, ml, mm, mr, bl, bm, br) =>
-        val tile = TileWithNeighbors(mm, Some(NeighboringTiles(tl, tm, tr, ml, mr,bl, bm, br))).withBuffer(buffer)
-        val extent = CogUtils.tmsLevels(z).mapTransform.keyToExtent(x, y)
+        fetchTile(self.uri.toString, z, xCoord, yCoord)
```
moradology: We may as well lean on GT contrib to do as much here as possible

pomadchin:
Won't work until https://github.com/geotrellis/maml/pull/70 is published btw

PR #62, opened 2018-12-07
Remove Encoder[Param] requirement in TMS

jisantuc:
Overview
-----

This PR removes the requirement that `Param`s have `Encoder` instances from `LayerTms`. It was inconvenient because it made it so that `fs2.Stream`s weren't valid params, which is a drag, because `fs2.Stream`s are the best.

Testing
-----

Compilation only


jisantuc:
...I didn't realize that this was removed in the previous `Encoder[Param]` removal. Nevermind

PR #74, opened 2019-01-15
Abstract over effect types

moradology:
Instead of using `IO`, we can use an `F` that satisfies the appropriate type classes.


moradology:
This is stale but could be useful later... closing for now

PR #80, opened 2019-02-18
Feature/wms map algebra

moradology:
This PR, with the help of MAML, adds the ability to specify complex layers for WMS consumption. As long as provided MAML expressions' variable names correspond to simple WMS layer names, the trees should be evaluable. For an example of what this looks like, pay special attention to `application.conf` and, in particular, the JSON provided under the heading of `algebra`

If misconfigured, the server should complain and crash before serving any requests


moradology:
superceded by #81

PR #81, opened 2019-02-25
Implement map algebra in WMS and reorganize for rest of OGC

moradology:
This PR implements support for map algebra within WMS services. A few distinctions introduced in this PR are worth calling out:
  * `OgcSource` vs `OgcLayer`: A source and a layer are not the same thing: sources define the location of imagery but do not contain sufficient information to render imagery in OGC services. An `OgcSource` + a `WmsRequest` are sufficient to produce an `OgcLayer` which has enough information to actually render imagery
  * `Simple` vs `MapAlgebra` Sources/Layers: Simple layers are merely imagery references whereas complex layers are 'complex' in that MAML expressions use the aforementioned references to build up map algebra programs which ultimately provide the desired imagery.

The flow of the server is now thus:
  1. On start, the conf is loaded. Simple layers are registered and complex layers validate that the necessary simple layers are defined
  2. XML interactions occur on the basis of provided layers. Clients use the XML communication to produce imagery requests
  3. Imagery requests are joined with `OgcSource`s to produce `OgcLayer`s which are rendered as appropriate


ogc/src/main/scala/geotrellis/server/ogc/OgcLayer.scala
```diff
@@ -0,0 +1,51 @@
...
+  source: RasterSource,
+  style: Option[StyleModel]
+) extends OgcLayer
+
+object SimpleLayer {
+  implicit val mapAlgebraLayerReification = new ExtentReification[SimpleLayer] {
+    def kind(self: SimpleLayer): MamlKind = MamlKind.Image
+    def extentReification(self: SimpleLayer)(implicit contextShift: ContextShift[IO]): (Extent, CellSize) => IO[Literal] =
+      (extent: Extent, cs: CellSize) =>  IO {
+        self.source.reprojectToGrid(self.crs, RasterExtent(extent, cs)).read(extent).map(RasterLit(_)).get
```
moradology: Is reproject to grid necessary here? What is it providing?

moradology: Using `reproject` in place of `reprojectToGrid` seemed to cause some memory-related issues (blowing the heap on the same layers that otherwise render appropriately)

PR #83, opened 2019-03-04
Add WMS Service Metadata Configuration

echeipesh:
This approach uses ScalaXB generated case classes to parse them using pureconfig.
While it may lead to some oddities, like requiring inclusion of `online-resource {}` property (which has default constructor) it maxims the configuration and reduces duplication.

The configuration object is threaded to the service to be returned on requests.

Sample configuration:

```
wms {
    service-metadata {
        name = "WMS"
        title = "GeoTrellis Service"
        online-resource {}
        keyword-list = {
            keyword = ["geotrellis", "catalog"]
        }
        contact-information {
            contact-person-primary {
                contact-person = "Eugene Cheipesh"
                contact-organization = "Azavea"
            }
            contact-position = "Developer"
            contact-address = {
                address-type = "Office"
                address = "990 Spring Garden St."
                city = "Philadelphia"
                state-or-province = "PA",
                post-code = "19087",
                country = "USA")
            }
        }
    }
}
```

### DIFF
--- a/ogc/src/main/scala/geotrellis/server/ogc/conf/Conf.scala
+++ b/ogc/src/main/scala/geotrellis/server/ogc/conf/Conf.scala
@@ -2,9 +2,16 @@ package geotrellis.server.ogc.conf
 
 import java.net.{InetAddress, URL}
 
+import geotrellis.server.ogc.wms.wmsScope
+import pureconfig.ConfigReader
+import scalaxb.DataRecord
+
+case class WMS(serviceMetadata: opengis.wms.Service)
+
 case class Conf(
   http: Conf.Http,
   service: Conf.Service,
+  wms: WMS,
   layers: List[OgcSourceConf]
 ) {
     def serviceUrl: URL = {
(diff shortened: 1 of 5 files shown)


ogc/src/main/scala/geotrellis/server/ogc/conf/Conf.scala
```diff
@@ -30,4 +37,27 @@ object Conf {
 
   lazy val conf: Conf = pureconfig.loadConfigOrThrow[Conf]
   implicit def ConfObjectToClass(obj: Conf.type): Conf = conf
+
+  implicit def nameConfigReader: ConfigReader[opengis.wms.Name] =
```
moradology: I think it is slightly more idiomatic to piggyback off of previously defined readers a la:
```scala  implicit val expressionReader: ConfigReader[Expression] =
    ConfigReader[String].map { expressionString =>
      decode[Expression](expressionString) match {
        case Right(success) =>
          success
        case Left(err) =>
          throw err
      }
    }```

File: ogc/src/main/scala/geotrellis/server/ogc/conf/Conf.scala [hunk 2]
```diff
@@ -39,13 +39,9 @@ object Conf {
...
-    ConfigReader.fromCursor[opengis.wms.Name] { cur =>
-    for {
-      str <- cur.asString.right
-    } yield {
+    ConfigReader[String].map { str =>
       opengis.wms.Name.fromString(str, wmsScope)
     }
-  }
 
   implicit def keywordConfigReader: ConfigReader[opengis.wms.Keyword] =
```
moradology: maybe here too

PR #91, opened 2019-03-19
Add WMTS tilematrixlimits support and cleanup all imports

pomadchin:
GetCapabilities that was generated: https://gist.github.com/pomadchin/4c2a63dd49f7388796a4a3cc3759dde4

### DIFF
--- a/ogc/src/main/scala/geotrellis/server/ogc/wmts/CapabilitiesView.scala
+++ b/ogc/src/main/scala/geotrellis/server/ogc/wmts/CapabilitiesView.scala
@@ -160,39 +160,55 @@ object CapabilitiesView {
     )
 
   implicit class OgcSourceMethods(val self: OgcSource) {
-    def toLayerType(layerName: String): LayerType = {
+    def toLayerType(layerName: String, tileMatrixSet: List[GeotrellisTileMatrixSet]): LayerType = {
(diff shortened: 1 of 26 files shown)


ogc/src/main/scala/geotrellis/server/ogc/wmts/CapabilitiesView.scala
```diff
@@ -160,39 +160,53 @@ object CapabilitiesView {
     )
 
   implicit class OgcSourceMethods(val self: OgcSource) {
-    def toLayerType(layerName: String): LayerType = {
+    def toLayerType(layerName: String, tileMatrixSet: List[GeotrellisTileMatrixSet]): LayerType = {
       val wgs84extent: Extent = ReprojectRasterExtent(self.nativeRE, self.nativeCrs.head, LatLng).extent
+      val tileMatrixLimits: List[TileMatrixLimits] = tileMatrixSet.flatMap(_.tileMatrix.map { tm =>
+        val gridBounds = tm.layout.mapTransform(self.nativeExtent)
```
echeipesh: This part seems suspect. I understand `OgcSource.nativeExtent` to be in the projection of the stored `RasterSource` while `TileMatrixSet` can be in arbitrary projection.

pomadchin: yes, right; so what projection it should be?

PR #92, opened 2019-03-20
Use trait-based styling

moradology:
Allowing for alternative styling implementations is most easily accomplished if we use a trait to unify different strategies. That's the path taken here.

Also, `StyleModel` is now `OgcStyle` because it wasn't really an M in MVC terms


ogc/src/main/scala/geotrellis/server/ogc/OgcStyle.scala
```diff
@@ -0,0 +1,55 @@
+package geotrellis.server.ogc
+
+import geotrellis.contrib.vlm.RasterSource
+
+import geotrellis.raster._
+import geotrellis.raster.histogram.Histogram
+import geotrellis.raster.render.{ColorMap, ColorRamp}
+
+trait OgcStyle {
```
jisantuc: :heart:

ogc/src/main/scala/geotrellis/server/ogc/wms/WmsService.scala
```diff
@@ -49,6 +49,7 @@ class WmsService(
 
   def routes: HttpRoutes[IO] = HttpRoutes.of[IO] {
     case req @ GET -> Root =>
+     println(s"""Recv'd request: $req""")
```
jisantuc: logger

PR #93, opened 2019-03-21
Add WMS LegendURL support

pomadchin:


ogc/src/main/resources/application.conf
```diff
@@ -293,6 +293,17 @@ layers = {
...
                 colors = ${color-ramps.red-to-blue}
                 stops = 64
+                legends = [
+                    {
+                        format = "image/png"
+                        width = 16
+                        height = 16
+                        online-resource = {
+                            type = "simple"
+                            href = "http://services.ga.gov.au/gis/rest/directories/capabilities/Northern_Australia_Land_Tenure/No_Value_Legend.png"
```
pomadchin: It's smth random from the internet btw.

PR #94, opened 2019-03-22
Update http4s and circe versions

jisantuc:
Overview
-----

This PR upgrades http4s and circe.

I upgraded http4s because 0.19.0 was EOL basically immediately, and 0.20 development is ongoing without any consideration for compatibility with 0.19.0 (and 0.19.0 also pins some typelevel deps pretty far behind current). I upgraded circe for consistency with some downstream consumers and nothing breaking after update and compilation.

Testing
-----

 * ci should be fine

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -2,10 +2,10 @@ import sbt._
 
 object Dependencies {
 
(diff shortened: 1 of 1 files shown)


moradology:
+1; merging

PR #102, opened 2019-04-09
Unify OGC services, wrapped in a CLI

moradology:
### The CLI
Currently, interaction with typesafe config is the primary way to control deployment behavior of the application. This is annoying for a couple of reasons: 
1. it isn't altogether clear how *highly* complex HOCON configurations (which we certainly depend on) ought to be specified at runtime
2. deployment configuration and application definition configuration are handled in the same place, meaning that reuse of configurations is not a straightforward affair (which is annoying and which testing more difficult)

Instead of the above, a thin CLI has been wrapped around the http4s server which takes three arguments:
1. local interface to bind http to
2. local port to bind http to
3. the URL to advertise services at (necessary for `GetCapabilities`)
4. the path to a HOCON configuration which defines layers/styling for OGC services (this will fallback to `application.conf` for testing/example purposes)

Screen Shot 2019-04-09 at 2 40 31 PM

### The unification of WMS/WMTS/WCS
Because `service` is a query parameter, we can avoid error-prone string/URL manipulation and host all services on the same public URL.


ogc-example/src/main/scala/geotrellis/server/ogc/conf/Conf.scala
```diff
@@ -11,15 +11,14 @@ import pureconfig.ConfigReader
...
-  http: Conf.Http,
   layers: Map[String, OgcSourceConf],
   wms: Conf.WMS,
   wmts: Conf.WMTS,
   wcs: Conf.WCS
 )
 
 object Conf {
-  trait OgcService {
+  trait OgcServiceConf {
```
pomadchin: Btw why these classes are inside of the object? If users would use them it would be really not convenient to call them.

ogc-example/src/main/scala/geotrellis/server/ogc/wcs/WcsView.scala
```diff
@@ -0,0 +1,78 @@
...
+            for {
+              describeCoverage <- IO { Operations.describeCoverage(wcsModel, p) }.attempt
+              result <- handleError(describeCoverage)
+            } yield {
+              logger.debug("describecoverage result", result)
+              result
+            }
+
+          case p: GetCoverageWcsParams =>
+            logger.debug(s"\033[1mGetCoverage: ${req.uri}\033[0m")
```
pomadchin: You can use ANSI COLORS library there

ogc-example/src/main/scala/geotrellis/server/ogc/wmts/WmtsView.scala
```diff
@@ -90,11 +82,7 @@ class WmtsService(
                 logger.error(err.toString, err)
                 InternalServerError(err.toString)
             }
-          }).getOrElse(BadRequest("No such layer"))
+          }).getOrElse(BadRequest("Layer not found"))
```
pomadchin: Hm, It's not critical, but mb it makes sense to add layer name into the BadRequest response?

ogc-example/src/main/scala/geotrellis/server/ogc/wcs/WcsView.scala
moradology: 😎

PR #109, opened 2019-05-03
Add llextents per WMS layer; properly escape JSON strings

moradology:
This PR adds lat/lng extents on each WMS layer and escapes JSON strings in the example configuration (correcting an oversight from bumping to `PureConfig` 0.10.2)

### DIFF
--- a/ogc-example/src/main/resources/application.conf
+++ b/ogc-example/src/main/resources/application.conf
@@ -256,7 +256,19 @@ layers = {
         type = "mapalgebrasourceconf"
         name = "addition-house-income"
         title = "test addition"
-        algebra = "{\"args\":[{\"name\":\"us-census-median-household-income\",\"symbol\":\"rasterV\"},{\"name\":\"us-census-median-household-income\",\"symbol\":\"rasterV\"}],\"symbol\":\"+\"}"
+        algebra = {
+          "args" : [
+            {
+              "name" : "us-census-median-household-income",
+              "symbol" : "rasterV"
+            },
+            {
+              "name" : "us-census-median-household-income",
+              "symbol" : "rasterV"
+            }
+          ],
+          "symbol" : "+"
+        }
         styles = [
             {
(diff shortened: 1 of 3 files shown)


ogc-example/src/main/resources/application.conf
```diff
@@ -256,19 +256,7 @@ layers = {
...
-              "symbol" : "rasterV"
-            },
-            {
-              "name" : "us-census-median-household-income",
-              "symbol" : "rasterV"
-            }
-          ],
-          "symbol" : "+"
-        }
+        algebra = "{\"args\":[{\"name\":\"us-census-median-household-income\",\"symbol\":\"rasterV\"},{\"name\":\"us-census-median-household-income\",\"symbol\":\"rasterV\"}],\"symbol\":\"+\"}"
```
pomadchin: Looks really bad. Mb you can write a manual decoder? Should not be too bad.

moradology: Agreed. I'm working on a JSON reader atm

ogc-example/src/main/resources/application.conf
```diff
@@ -51,7 +51,7 @@ wcs = {
             description = "Geotrellis WCS Service"
             keywords = []
             profile = ["http://azavea.com/wcs-profile"]
-            fees = []
+            fees = ""
```
pomadchin: ah no, should it be reverted?

moradology: That field is not currently being used and i cut a release to test with the fixes included here - this change is a fix to compensate for a purescript upgrade. I'm investigating right now to determine how the decoder ought to change

moradology: String is a better representation after all. It looks like `[]` was being parsed to `None` whereas purescript treats missing fields as `None`s
image

pomadchin: wow gotcha

PR #116, opened 2019-05-28
Upgrade MAML version

aaronxsu:
This PR upgrades the MAML version to `0.4.0` and fixes some compilation errors as a result of upgrading. 

`ogc/compile`, `core/compile`, and `example/compile` should succeed with no errors.

Closes https://github.com/azavea/raster-foundry-platform/issues/727

### DIFF
--- a/core/src/main/scala/geotrellis/server/LayerExtent.scala
+++ b/core/src/main/scala/geotrellis/server/LayerExtent.scala
@@ -12,6 +12,7 @@ import com.typesafe.scalalogging.LazyLogging
 import io.circe._
 import io.circe.syntax._
 import cats._
+import cats.data.Validated._
 import cats.data.{NonEmptyList => NEL}
 import cats.effect._
 import cats.implicits._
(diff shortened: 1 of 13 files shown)


core/src/main/scala/geotrellis/server/LayerExtent.scala
```diff
@@ -78,11 +85,11 @@ object LayerExtent extends LazyLogging {
     param: Param
   )(
     implicit reify: ExtentReification[Param],
-             contextShift: ContextShift[IO]
+             contextShift: ContextShift[IO],
+             applicativeError: ApplicativeError[IO, NEL[MamlError]]
   ): (Extent, CellSize) => IO[Interpreted[MultibandTile]] =
     (extent: Extent, cellsize: CellSize) => {
-      val eval = curried(RasterVar("identity"), Interpreter.DEFAULT)
+      val eval = curried(RasterVar("identity"), ConcurrentInterpreter.DEFAULT)
```
jisantuc: Worth calling out -- we explicitly shifted to the `ConcurrentInterpreter` everywhere in gt-server, since all of the typeclasses use `IO` as an effect type, and the `NaiveInterpreter` doesn't know anything about `IO`

moradology: Do we have a sense of what the performance impacts might be? I'm assuming negligible, at worst, but it would be nice to know for sure

jisantuc: Should be positive but we don't have load tests here :man_shrugging:

PR #125, opened 2019-06-11
[WIP] Fix OGC Tile Clipping

notthatbreezy:
This is a first crack at trying to fix the OGC tile clipping. I'm not sure if this is the 100 percent correct approach, but it seems to solve the problem locally when I use it.

I'm unsure how to test this more fully though to ensure that this isn't breaking some other mode of WMS.

Note: There is a first commit that has changes necessary to make the OGC example project runnable again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,10 +16,12 @@ serveNdviExample:
 serveGdalNdviExample:
 	docker-compose run --service-ports gdal-ndvi-example
 
+serveOgcExample:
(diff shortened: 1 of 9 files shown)


notthatbreezy:
Closing this - the bug turned out to be in [`MosaicRasterSource`](https://github.com/geotrellis/geotrellis-contrib/pull/202)

PR #132, opened 2019-07-02
Use 2.12 compatible ansi-interpolator

moradology:
This works with 2.12 as well as 2.11; the former version did not


moradology:
Fairly trivial; merging

PR #143, opened 2019-08-23
Include opengis in `root` project

notthatbreezy:
## Overview

Adds `opengis` to the `root` project to prevent some head-scratching when publishing local for development.

### Checklist

- [ ] Description of PR is in an appropriate section of the CHANGELOG and grouped with similar changes if possible

## Testing Instructions

- Start an sbt shell, run `publishLocal` and see that artifacts for `opengis` are also published


notthatbreezy:
Looks like this has been fixed by a different PR :)

PR #145, opened 2019-09-06
Update STAC datamodel for label extension support

### LINKED ISSUE #141: Update STAC link type and allow GeoJSON asset for STAC Item
This issue is to request two modifications to the STAC related types and data model:
1. Add a `Source` type to `StacLinkType`, which is proposed [here in STAC label extension spec](https://github.com/radiantearth/stac-spec/tree/dev/extensions/label#links-source-imagery), and which is used in the [STAC export builder](https://github.com/raster-foundry/raster-foundry/pull/5091/files#diff-e7e313dd0579d92f5777d054888b5989R222) (it is `VendorLinkType("Source")` for now).
2. We'd like `StacItem` to support a `StacAsset` of type `application/geo+json` so that a STAC item can have a geojson asset as its label data in a GeoJSON Feature Collection. ([see proposed spec here](https://github.com/radiantearth/stac-spec/tree/dev/extensions/label#assets))

jisantuc:
## Overview

This PR relaxes the requirements on `StacItem`s so that they can be created without a COG asset and
adds the `Source` link type to `StacLinkTypes`. The `Source` link type is a part of the labeling extension, which is [included in STAC 0.8.0-rc1](https://github.com/radiantearth/stac-spec/blob/v0.8.0-rc1/CHANGELOG.md#added), and the relaxes requirement is part of making the datamodel usable for any purpose other than the demo that originally consumed it.

### Checklist

## Testing Instructions

- ~ci~ i saw a `.circleci` dir and thought that it meant there's ci. oops.
- `stac/test` -- shows that the new link type is fine
- `example/compile` shows that no type contracts were broken through the example

Closes #141

### DIFF
--- a/stac/src/main/scala/StacLink.scala
+++ b/stac/src/main/scala/StacLink.scala
@@ -8,16 +8,18 @@ case class StacLink(
     href: String,
     rel: StacLinkType,
     _type: Option[StacMediaType],
-    title: Option[String]
+    title: Option[String],
+    labelExtAssets: List[String]
 )
 
 object StacLink {
-  implicit val encStacLink: Encoder[StacLink] = Encoder.forProduct4(
+  implicit val encStacLink: Encoder[StacLink] = Encoder.forProduct5(
     "href",
     "rel",
     "type",
-    "title"
-  )(link => (link.href, link.rel, link._type, link.title))
+    "title",
+    "label:assets"
+  )(link => (link.href, link.rel, link._type, link.title, link.labelExtAssets))
 
   implicit val decStacLink: Decoder[StacLink] = new Decoder[StacLink] {
     final def apply(c: HCursor) =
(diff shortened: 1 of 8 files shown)


stac/src/main/scala/StacLink.scala
```diff
@@ -8,24 +8,27 @@ case class StacLink(
     href: String,
     rel: StacLinkType,
     _type: Option[StacMediaType],
-    title: Option[String]
+    title: Option[String],
+    labelExtAssets: List[String]
```
aaronxsu: Do you think we probably should make this field `labelExtAssets: Option[List[String]]`?

aaronxsu: Update: based on an off-line discussion, `labelExtAssets: List[String]` should suffice at this stage.

Issue #72, opened 2019-01-10
Parameterize histogram generation w/ sampling strategy


moradology:

echeipesh:
Adding a bit of context here. This issue came up again and @pomadchin and I discussed the idea of having `Histogram` be a MAML expression. This would allow the interpreter to do the job of figuring out how to best extract a histogram from the parameter.

PR #167, opened 2019-10-25
Make LayerHistogram more tolerant of param lists

### LINKED ISSUE #165: Bug in LayerHistogram for Params with Multiple Extents
This issue is based on a bug we just ran into in RF where [`LayerHistogram`](https://github.com/geotrellis/geotrellis-server/blob/develop/core/src/main/scala/geotrellis/server/LayerHistogram.scala#L41-L44) suspiciously throws an error when we wouldn't expect it.

The lines linked above seems to be the crux of the issue - in our case our params look something like `Map("RedBand" -> (rs1, rs2, rs3), "GreenBand" -> (rs1, rs2, rs3))` and the flattening happens such that we end up with a list like `List(extentRs1, extentRs2, extentRs3)` which do not overlap (or if they do very minimally and an exception ends up getting thrown.

@jisantuc is that a good description of the problem?

I think a solution might end up looking like:

```scala
NEL.fromListUnsafe(params.values.toList)
  .traverse(_.rasterExtents.map(_.reduce))
```

I'm not sure if there is a semigroup available for `RasterExtent` somewhere, but that's all that would be needed for that to compile.

notthatbreezy:
## Overview

This solves a bug where params that had lists of values
themselves could end up falsely saying that their raster
extents did not intersect (even though we're 100 percent
sure they do) and then failing to produce a
histogram for the application.

### Checklist

### Demo

image

### Notes

I think this is another example of where https://github.com/geotrellis/geotrellis-server/issues/72 being closed would have given us a few escape hatches, but ultimately this still was probably a bug here.

## Testing Instructions

Unfortunately I don't have good testing instructions outside of a Raster Foundry environment. But once this branch is published local and you've updated your dependencies in RF this is testable by viewing/creating any multi-scene analysis.

Closes #165

### DIFF
--- a/core/src/main/scala/geotrellis/server/LayerHistogram.scala
+++ b/core/src/main/scala/geotrellis/server/LayerHistogram.scala
@@ -25,74 +25,104 @@ object LayerHistogram extends LazyLogging {
   case class NoSuitableHistogramResolution(cells: Int) extends Throwable
   case class RequireIntersectingSources() extends Throwable
 
+  // Added so that we can get combine
+  implicit val extentSemigroup: Semigroup[Extent] =
+    new Semigroup[Extent] {
+      def combine(x: Extent, y: Extent): Extent = x.combine(y)
+    }
+
   // Provide IOs for both expression and params, get back a tile
   def apply[Param](
-    getExpression: IO[Expression],
-    getParams: IO[Map[String, Param]],
-    interpreter: Interpreter[IO],
-    maxCells: Int
+      getExpression: IO[Expression],
+      getParams: IO[Map[String, Param]],
+      interpreter: Interpreter[IO],
+      maxCells: Int
   )(
-    implicit reify: ExtentReification[Param],
-             extended: HasRasterExtents[Param],
-             contextShift: ContextShift[IO]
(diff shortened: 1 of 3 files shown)


core/src/main/scala/geotrellis/server/LayerHistogram.scala
```diff
@@ -25,74 +25,104 @@ object LayerHistogram extends LazyLogging {
...
-      _                 <- IO { logger.debug(s"[LayerHistogram] Largest cell size of provided layers calculated: $cellSize") }
-      mbtileForExtent   <- IO { LayerExtent(getExpression, getParams, interpreter) }
-      _                 <- IO { logger.debug(s"[LayerHistogram] calculating histogram from (approximately) ${intersection.area / (cellSize.width * cellSize.height)} cells") }
-      interpretedTile   <- mbtileForExtent(intersection, cellSize)
+      params <- getParams
+      rasterExtents <- NEL
+        .fromListUnsafe(params.values.toList)
+        .traverse(_.rasterExtents)
+        .map(_.flatten)
+      extents <- NEL
```
notthatbreezy: This is the relevant change - the rest is `scalafmt`

PR #170, opened 2019-10-29
Use Refined to discriminate between Public/Private Stac Collections

notthatbreezy:
## Overview

This commit refactors StacCollection to be an ADT with two
implementations:

 1. A public STAC collection that has a valid SPDX identifier as its
 license
 2. A proprietary STAC collection that requires a link to a license in
 its list of links

### Checklist

- [ ] Description of PR is in an appropriate section of the CHANGELOG and grouped with similar changes if possible

### Notes

I don't want to merge this until I have a chance to publish local and use it to resolve https://github.com/raster-foundry/raster-foundry/issues/5212

I could use some guidance on what tests we want to see for this - I guess for the most part checking that the predicates I wrote actually work are the ones that should be written.

## Testing Instructions

- Review organization and code to see if it makes sense

Partially Closes https://github.com/raster-foundry/raster-foundry/issues/5212

### DIFF
--- a/stac/src/main/scala/StacLicense.scala
+++ b/stac/src/main/scala/StacLicense.scala
@@ -1,0 +1,41 @@
+package geotrellis.server.stac
+
+import cats.implicits._
+import io.circe._
+
+sealed trait StacLicense {
+  val name: String
+}
+
+final case class Proprietary() extends StacLicense {
+  val name: String = "proprietary"
+}
+
+final case class SPDX(spdxId: SpdxId) extends StacLicense {
+  val name = spdxId.value
+}
+
+object StacLicense {
+  implicit val encoderStacLicense: Encoder[StacLicense] =
+    Encoder.encodeString.contramap[StacLicense](_.name)
+
+  implicit val decodeSpdx: Decoder[SPDX] = Decoder.decodeString.emap { s =>
+    SpdxId.from(s) match {
+      case Left(error) => Either.left(error)
+      case Right(spdx) => Either.right(SPDX(spdx))
+    }
+  }
+
+  implicit val decodeProprietary: Decoder[Proprietary] =
+    Decoder.decodeString.emap {
+      case "proprietary" => Either.right(Proprietary())
+      case s             => Either.left(s"Unknown License: $s")
+    }
+
+  implicit val decodeStacLicense: Decoder[StacLicense] =
+    List[Decoder[StacLicense]](
+      Decoder[Proprietary].widen,
+      Decoder[SPDX].widen
+    ).reduceLeft(_ or _)
+
+}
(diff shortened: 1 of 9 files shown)


stac/src/main/scala/package.scala
```diff
@@ -1,105 +1,157 @@
...
 import io.circe.parser.{decode, parse}
 import io.circe.shapes.CoproductInstances
 import shapeless._
 
-import java.text.ParseException
-import java.time.Instant
-
 package object stac {
+
+  case class LicenseLink()
```
notthatbreezy: Not sure where I should put these, so for now they are here

stac/src/main/scala/package.scala
```diff
@@ -1,105 +1,157 @@
...
+            case _       => false
+          },
+        t => s"Not a License Link: ${t.rel}",
+        LicenseLink()
+      )
+  }
+
+  type StacLinksWithLicense = List[StacLink] Refined Exists[LicenseLink]
+  object StacLinksWithLicense
+      extends RefinedTypeOps[StacLinksWithLicense, List[StacLink]]
```
notthatbreezy: I didn't know this until last night, but using type aliases in this way lets us do something like:

```scala
val result: Either[Left, Right(StacLinksWithLicense)] = StacLinksWithLicense.from(x: List[StacLink])
```

which I think is a totally acceptable requirement for consumers of this library to deal with

stac/src/main/scala/StacLicense.scala
```diff
@@ -0,0 +1,41 @@
+package geotrellis.server.stac
+
+import cats.implicits._
+import io.circe._
+
+sealed trait StacLicense {
+  val name: String
+}
+
+final case class Proprietary() extends StacLicense {
```
moradology: Perhaps just a `case object` if non-parametric creation

stac/src/main/scala/package.scala
moradology: Sounds reasonable to me. It also looks like `RefinedTypeOps` provides an escape hatch for runtime checks, allowing `StacLinksWithLicense.apply(x: List[StacLink])`.

moradology: Nevermind. Looks like the apply is available via macros and you've got to call `unsafeFrom` to avoid compile-time checks

stac/src/main/scala/StacLicense.scala
jisantuc: I'm slightly surprised this works

```scala
@ trait Foo {
    val x: Int
  } 
defined trait Foo

@ case object Bar extends Foo {
    val x: Int = 3
  } 
defined object Bar
```

PR #206, opened 2020-02-17
Add support for colormap with floating point keys

### LINKED ISSUE #187: Unsupported non-integer value as ColorMap keys
When we create a color-map with a non-integer value, we get an exception message.

**Color-map example:**
"ndvi": {
    -1.0: 0x1947B0FF,
    -0.7: 0x3961B7FF,
    -0.5: 0x5A7BBFFF,
    -0.4: 0x7B95C6FF,
    -0.3: 0x9CB0CEFF,
    -0.2: 0xBDCAD5FF,
    -0.1: 0xDEE4DDFF,
    0: 0xFFFFE5FF,
    0.1: 0xDAE4CAFF,
    0.2: 0xB6C9AFFF,
    0.3: 0x91AF94FF,
    0.4: 0x6D9479FF,
    0.5: 0x487A5EFF,
    0.7: 0x245F43FF,
    1.0: 0x004529FF
}

### Exception message:
Exception in thread "main" pureconfig.error.ConfigReaderException: Cannot convert configuration to a geotrellis.server.ogc.conf.Conf. Failures are:
  at 'layers.2014-ndvi.styles.0.colors.0':
    - (file:/opt/datacube/applications/application.conf:732) Expected type STRING. Found OBJECT instead.
  at 'layers.2014-ndvi.styles.0.colors.1':
    - (file:/opt/datacube/applications/application.conf:740) Expected type STRING. Found OBJECT instead.
  at 'layers.2014-ndvi.styles.0.colors.2':
    - (file:/opt/datacube/applications/application.conf:746) Expected type STRING. Found OBJECT instead.
  at 'layers.2015-ndvi.styles.0.colors.0':
    - (file:/opt/datacube/applications/application.conf:732) Expected type STRING. Found OBJECT instead.

moradology:
## Overview

This PR adds support for `ColorMap` with non-integer keys by looking for an incorrectly parsed value. If incorrectly parsed, the non-integer key is reconstructed. This appears to be safe because this code path is only exercised when we're attempting to decode a `ColorMap` and not more generally as HOCON is being parsed

### Checklist

### Demo

The example configuration now includes a set of floating point keyed `ColorMap`s

## Testing Instructions

- Run the example OGC server

Closes #187

### DIFF
--- a/ogc-example/src/main/resources/application.conf
+++ b/ogc-example/src/main/resources/application.conf
@@ -222,6 +222,12 @@ layers = {
                 type = "colorrampconf"
                 colors = ${color-ramps.red-to-blue}
                 stops = 64
+            },
+            {
+                name = "ndvi"
+                title = "NDVI colors (for demonstration only)"
+                type = "colormapconf"
+                color-map = ${color-maps.ndvi}
             }
         ]
     }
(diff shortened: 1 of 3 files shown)


ogc-example/src/main/resources/application.conf
```diff
@@ -369,4 +375,21 @@ color-maps = {
     2500: 0xD7D7D7FF,
     3000: 0xFFFFFFFF
   }
+  "ndvi": {
+    -1.0: 0x1947B0FF,
```
pomadchin: btw, is not it better just to have strings everywhere if doubles are used? just for simplicity?

moradology: Whose simplicity? I'm open to coming down on a best practice hard if it is obviously useful but am concerned that hard+fast rules about quoting in some contexts rather than others will make user's lives a bit more difficult (as, generally, HOCON seems to be aimed at a very permissive syntax)

PR #207, opened 2020-02-17
QueryDSL

### LINKED ISSUE #162: RasterSource Catalog
We wish to be able to query a spatiotemporal "catalog" using `RasterSource` interface. This could be a GeoTrellis layer with temporal dimensions or it could be a STAC catalog where there are multiple COGs (Cloud Optimized GeoTiffs) available with differing time stamps.

In either case we are currently limited because the `RasterSource` interface presents a single raster both in its metadata and in its read method interfaces.

It is not desirable to further complicate the `RasterSource` interface by adding temporal query capability. Further it is likely that different time slices have may be spread over multiple files. Switching source of IO and potentially reading metadata should not be hidden from user of `RasterSource` interface at those are potentially expensive and error-prone operations that must be explicit and carefully considered by the user.

Instead we should preserve `RasterSource` as a "spatial slice" of otherwise potentially multi-dimensional dataset. Therefore "something" should be able to produce `RasterSource` instances given a spatiotemporal query. Let's call this something `RasterCatalog` because it maps nicely onto STAC catalog.

At a minimum:
```scala
trait RasterCatalog {
  def find(query: Query): Seq[RasterSource]
}
```

Since `RasterSource` as an interface is supposed to be lazy we can return large number of them safely.Additionally if the query source (like STAC catalog) had record the raster metadata, it could be included in a specialized wrapper, further delaying the initial header read until the first `RasterSource.read()` call is made.

`Query` type should encapsulate the query as an ADT, so it can inspected, serialized and optimized before actual execution.

Critically this interface should avoid making any assumptions about structure of the data because it can commonly take many forms. 

For instance:

- GeoTrellis spatiotemporal layer/pyramid
- STAC catalog
- Folder of COGs with timestamp encoded in filename

### LINKED ISSUE #163: {WCS|WMTS|WMS}Model uses RasterSource catalog
Currently `OgcModel`s assume that all entries are a simple `OgcSource` which can be either `MapAlgebraSource` or `SimpleOgcLayer`.

`SimpleOgcLayer` is bound to `RasterSource` instance and therefore can not respond to temporal query.

This model should be updated to make use of the new [`RasterCatalog`](https://github.com/geotrellis/geotrellis-server/issues/162) such that the WCS/WMS/WMTS query parameters and their temporal component can be evaluated to produce correct instance of `RasterSource`s for given time stamp.

If more than one are returned the service should render raster from the source with the earliest timestamp. This implies that information about timestamp should be easily available to the `WcsModel`.

### LINKED ISSUE #204: RasterSourceCatalog query DSL
The task is to implement a DSL that will allow to query [the RasterCatalog](https://github.com/geotrellis/geotrellis-server/issues/162).

The usual pipeline should be:

1. parse the input schema specific (WMS, WCS, WMTS) parameters and convert them into the query (`ana` or `unfold`)
2. Evaluate the query and get the query result (`cata` or `foldRight`)

pomadchin:
## Overview

This PR adds a concept of RasterSource catalog that is possible to query via they special query DSL

### Checklist

Closes #162
Closes #163
Closes #204


core/src/main/scala/geotrellis/store/GeoTrellisRasterSourceLegacy.scala
```diff
@@ -17,44 +17,102 @@
...
+import geotrellis.raster.io.geotiff.{AutoHigherResolution, OverviewStrategy}
 import geotrellis.raster.resample.{NearestNeighbor, ResampleMethod}
 import geotrellis.raster._
 import geotrellis.layer._
+import geotrellis.layer.filter._
 import geotrellis.vector._
 
+import java.time.ZonedDateTime
+
+case class LayerLegacy(id: LayerId, metadata: TileLayerMetadata[_], bandCount: Int) {
```
pomadchin: Since we have GeoTrellisLayers here, why not to keep all the changes here and after approval port them into GeoTrellis core

core/src/main/scala/geotrellis/store/GeoTrellisRasterSourceLegacy.scala
```diff
@@ -17,44 +17,102 @@
...
+  def crs: CRS = layerMetadata.crs
+
+  def cellType: CellType = dstCellType.getOrElse(layerMetadata.cellType)
+
+  def attributes: Map[String, String] = Map(
+    "catalogURI" -> dataPath.value,
+    "layerName"  -> layerId.name,
+    "zoomLevel"  -> layerId.zoom.toString,
+    "bandCount"  -> bandCount.toString
+  ) ++ time.map(t => ("time", t.toString)).toMap
```
pomadchin: The idea is that we can add time into attributes in case it is a temporal layer

core/src/main/scala/geotrellis/store/GeoTrellisRasterSourceLegacy.scala
```diff
@@ -86,48 +144,86 @@ class GeoTrellisRasterSourceLegacy(
...
+    }
 
     val header = reader.attributeStore.readHeader[LayerHeader](layerId)
 
     if (!header.keyClass.contains("spark")) {
       (Class.forName(header.keyClass), Class.forName(header.valueClass)) match {
-        case (SpatialKeyClass, TileClass) => spatialTileRead
-        case (SpatialKeyClass, MultibandTileClass) => spatialMultibandTileRead
+        case (SpatialKeyClass, TileClass)            => spatialTileRead
+        case (SpatialKeyClass, MultibandTileClass)   => spatialMultibandTileRead
```
pomadchin: I definitely feel bad about GeoTrellisRasterSource

core/src/main/scala/geotrellis/store/query/QueryCollection.scala
```diff
@@ -0,0 +1,6 @@
+package geotrellis.store.query
+
+trait QueryCollection[T, G[_]] {
```
pomadchin: I tried to avoid the word 'catalog'; what about `RasterSourceRepository`?

ogc/src/main/scala/geotrellis/server/ogc/wcs/WcsParams.scala
```diff
@@ -65,6 +66,8 @@ case class GetCoverageWcsParams(
   gridOffsets: (Double, Double),
   crs: CRS
 ) extends WcsParams {
+  def toQuery: Query = withName(identifier) and intersects(ProjectedExtent(extent, crs))
```
pomadchin: It is a very good question, should the Params => Query function live inside a case class, or we can generate query during the JSON traverse and to perform a direct JSON => Query conversion.

ogc/src/main/scala/geotrellis/server/ogc/Render.scala
```diff
@@ -78,11 +78,9 @@ object Render {
...
       val (currentEdge, nextEdge) = breaks(counter) -> breaks(counter + 1)
       val step = (nextEdge - currentEdge) / lengthBetween
-      val points: List[Double] = if (step == 0) {
-        List.empty[Double]
-      } else {
-        (currentEdge to nextEdge by step).toList
-      }
+      val points: List[Double] =
+        if (step == 0) List.empty[Double]
+        else Range.BigDecimal.inclusive(currentEdge, nextEdge, step).map(_.toDouble).toList
```
pomadchin: @moradology ^ changed the syntax due to deprecation warnings

build.sbt
```diff
@@ -186,9 +186,19 @@ lazy val core = project
...
-      scalatest
+      scalatest,
+      droste
     )
   )
+  .settings(
+    libraryDependencies ++= {
+      CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((2, scalaMajor)) if scalaMajor == 11 =>
+          Seq(circeJava8.value)
```
pomadchin: In Scala 2.12 java8.time codecs are built in, that is why we have a compact object

ogc/src/main/scala/geotrellis/server/ogc/wcs/WcsParams.scala
pomadchin: I'll mark it as resolved, we can construct Query from the input Json now.

core/src/main/scala/geotrellis/store/query/package.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  implicit class RasterSourceOps(self: RasterSource) {
+    def projectedExtent: ProjectedExtent = ProjectedExtent(self.extent, self.crs)
+  }
+
+  implicit class ProjectedExtentOps(self: ProjectedExtent) {
+    def intersects(projectedExtent: ProjectedExtent): Boolean = self.extent.intersects(projectedExtent.reproject(self.crs))
+    def covers(projectedExtent: ProjectedExtent): Boolean     = self.extent.covers(projectedExtent.reproject(self.crs))
+    def contains(projectedExtent: ProjectedExtent): Boolean   = self.extent.contains(projectedExtent.reproject(self.crs))
+  }
```
echeipesh: @pomadchin think these are worth just moving to GT main after this? Seems a shame that we still don't have a way to abstract over both extent and ProjectedExtent but also a shame to have to write utility code in apps.

pomadchin: @echeipesh 💯 agree, along with all GeoTrellisRasterSources enhancements that are implemented here and probably with QueryDSL as well

ogc-example/src/main/scala/geotrellis/server/ogc/conf/OgcServiceConf.scala
```diff
@@ -28,12 +27,12 @@ import geotrellis.server.ogc.wmts.GeotrellisTileMatrixSet
  */
 sealed trait OgcServiceConf {
   def layerDefinitions: List[OgcSourceConf]
-  def layerSources(simpleSources: List[SimpleSource]): List[OgcSource] = {
+  def layerSources(simpleSources: List[SimpleSource]): OgcSourceRepository = {
     val simpleLayers =
-      layerDefinitions.collect { case ssc@SimpleSourceConf(_, _, _, _) => ssc.model }
+      layerDefinitions.collect { case ssc @ SimpleSourceConf(_, _, _, _) => ssc.models }.flatten
```
echeipesh: This looks like it may result in each layer being listed once per time-stamp in GetCapabilities ?

core/src/main/scala/geotrellis/store/GeoTrellisRasterSourceLegacy.scala
```diff
@@ -137,23 +235,75 @@ object GeoTrellisRasterSourceLegacy {
...
+        case ki if ki.isInstanceOf[ZSpaceTimeKeyIndex]       => ki.asInstanceOf[ZSpaceTimeKeyIndex].temporalResolution
+        case ki if ki.isInstanceOf[HilbertSpaceTimeKeyIndex] => ki.asInstanceOf[ZSpaceTimeKeyIndex].temporalResolution
+        case ki => throw new UnsupportedOperationException(s"Can't derive layer temporal resolution for ${ki.getClass}, try to pass a tResolution parameter explicitly.")
+      })
+
+      md.bounds match {
+        case KeyBounds(minKey, maxKey) =>
+          (minKey.time.toInstant.toEpochMilli to maxKey.time.toInstant.toEpochMilli by temporalResolution).toList.map { time =>
+            new GeoTrellisRasterSourceLegacy(attributeStore, dataPath, sourceLayers, Some(ZonedDateTime.ofInstant(time, ZoneOffset.UTC)), targetCellType)
+          }
```
echeipesh: This unfortunately is not correct for temporal layers. There could be time buckets that have no records stored in them and there could be time buckets which have multiple records stored in them. The only way to know which is which is to have use store the used values in metadata. Thinking of a concrete suggestion for this.

core/src/main/scala/geotrellis/store/GeoTrellisRasterSourceLegacy.scala
echeipesh: Because of this match or other reasons too ?

core/src/main/scala/geotrellis/store/GeoTrellisRasterSourceLegacy.scala
pomadchin: @echeipesh provide time as an input into this build function? (this can be a part of configuration) 

But what to do if you don't know anything about the temporal layer?

ogc/src/main/scala/geotrellis/server/ogc/wcs/GetCoverage.scala
```diff
@@ -52,15 +52,13 @@ class GetCoverage(wcsModel: WcsModel) {
...
-            LayerExtent.identity(SimpleOgcLayer(name, title, params.crs, source, None))
-          case MapAlgebraSource(name, title, sources, algebra, styles) =>
-            val simpleLayers = sources.mapValues { rs => SimpleOgcLayer(name, title, params.crs, rs, None) }
+
+        val eval = wcsModel.getLayers(params).headOption.map {
+          case so @ SimpleOgcLayer(_, _, _, _, _) => LayerExtent.identity(so)
+          case MapAlgebraOgcLayer(_, _, _, simpleLayers, algebra, _) =>
             LayerExtent(IO.pure(algebra), IO.pure(simpleLayers), ConcurrentInterpreter.DEFAULT)
-        }
+        }.getOrElse(throw new Exception("bad interpreter"))
```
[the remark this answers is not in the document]
pomadchin: I think you're correct; Originally it was just unsafe `.get` from a map & probably it deserves a separate issue to walk though all the exceptions we throw in the code of gt-server?

core/src/main/scala/geotrellis/store/GeoTrellisRasterSourceLegacy.scala
echeipesh: Can't really report anything you don't know. The alternative is that you report a date for which there is no data and query comes back empty.

core/src/main/scala/geotrellis/store/query/QueryF.scala
```diff
@@ -0,0 +1,100 @@
...
+  @JsonCodec case class Or[A](left: A, right: A) extends QueryF[A]
+  @JsonCodec case class And[A](left: A, right: A) extends QueryF[A]
+  @JsonCodec case class Intersects[A](projectedExtent: ProjectedExtent) extends QueryF[A]
+  @JsonCodec case class Contains[A](projectedExtent: ProjectedExtent) extends QueryF[A]
+  @JsonCodec case class Covers[A](projectedExtent: ProjectedExtent) extends QueryF[A]
+  @JsonCodec case class At[A](time: ZonedDateTime, fieldName: String = "time") extends QueryF[A]
+  @JsonCodec case class Between[A](from: ZonedDateTime, to: ZonedDateTime, fieldName: String = "time") extends QueryF[A]
+  @JsonCodec case class WithName[A](name: String) extends QueryF[A]
+  @JsonCodec case class WithNames[A](names: Set[String]) extends QueryF[A]
+  @JsonCodec case class Nothing[A]() extends QueryF[A]
```
moradology: Perhaps `None` is a better complement to `All`? `Nothing` seems to me like literally 'no thing' whereas this is just none of the instances of w.e particular thing we're talking about

pomadchin: @moradology `None` is a reserved word; and the `none` function potentially can confuse users. I'm happy to try Any / Nothing though but probably `any` can also confuse people as a part of a Query DSL

ogc-example/src/main/scala/geotrellis/server/ogc/conf/RasterSourceConf.scala
```diff
@@ -25,7 +25,7 @@ import geotrellis.store.{GeoTrellisPath, GeoTrellisRasterSourceLegacy}
  *  a geotrellis-contrib [[RasterSource]].
  */
 sealed trait RasterSourceConf {
-  def toRasterSource: RasterSource
+  def toRasterSources: List[RasterSource]
```
moradology: This seems like a significant alteration, but I couldn't figure out what it was for (the only implementation I saw here just attached the lone RS to `Nil`)

core/src/main/scala/geotrellis/store/query/QueryF.scala
moradology: :yoda:

ogc-example/src/main/scala/geotrellis/server/ogc/conf/RasterSourceConf.scala
pomadchin: It was to make use of the Repository interface. The whole thing was to make queriable temporal layers and a temporal layer is just a set of RasterSources. So having a List of RasterSources read from the configuration can allow as to create an instance of a Repository that we can query by time as well.

PR #208, opened 2020-02-18
Fix logging

### LINKED ISSUE #144: Log lib conflicts
Right now attempting to read from HDFS from Docker results in failure because it uses JCL and requires Log4j, which is currently excluded by the build strategy. I I fixed it by adding these exclusions:
```
    excludeDependencies ++= Seq(
      ExclusionRule("log4j", "log4j"),
      ExclusionRule("commons-logging", "commons-logging"),
      ExclusionRule("org.slf4j", "slf4j-log4j12"),
      ExclusionRule("org.slf4j", "slf4j-nop")
    )
```
And adding these dependencies to bridge all the loggers to SLF4J:
```
  "ch.qos.logback" % "logback-classic" % "1.2.3"
  "ch.qos.logback" % "logback-core" % "1.2.3"
  "org.slf4j" % "jcl-over-slf4j" % "1.7.25"
  "org.slf4j" % "log4j-over-slf4j" % "1.7.25"
  "org.slf4j" % "slf4j-api" % "1.7.25"
```

### LINKED ISSUE #205: Review logging behavior
Review the logging behavior and / or update the DOCS with instructions how to change logging level for the core part or for the examples (and ogc-examples) projects.

moradology:
## Overview

Logging stopped working at some point and this PR rectifies that situation so that we don't need to lean on `println`

### Checklist

### Demo

image

Run the main and watch logging appear

### Notes

This was a confusing issue. Resolution required excluding the `NOP` logging jar from SLF4J despite this apparently being a fallback logger which shouldn't have blocked SLF4J from binding correctly

Closes #205
Closes #144


ogc-example/src/main/resources/logback.xml
```diff
@@ -7,6 +7,7 @@
 
   <logger name="geotrellis.server.ogc" level="${GT_OGC_LOG_LEVEL:-INFO}"/>
   <logger name="org.http4s.blaze.channel.nio1" level="WARN"/>
+  <logger name="software.amazon.awssdk" level="WARN"/>
```
moradology: This lib can be a bit chatty, though the warnings can be useful

pomadchin: Can we make it also configurable? probably other global vars?

project/Dependencies.scala
```diff
@@ -92,16 +92,14 @@ object Dependencies {
...
   val macrosParadise = "org.scalamacros" % "paradise" % "2.1.0"
   val scalaParser = "org.scala-lang.modules" %% "scala-parser-combinators" % "1.1.1"
   val dispatch = "net.databinder.dispatch" %% "dispatch-core" % dispatchVer
   val apacheMath = "org.apache.commons" % "commons-math3" % "3.6.1"
   val commonsIo = "commons-io" % "commons-io" % "2.6"
-  val slf4jApi = "org.slf4j" % "slf4j-api" % "1.7.25"
-  val slf4jSimple = "org.slf4j" % "slf4j-simple" % "1.7.25"
-  val logback = "ch.qos.logback" % "logback-classic" % "1.1.7"
   val shapeless = "com.chuusai" %% "shapeless" % shapelessVer
+  val scalaLogging = "com.typesafe.scala-logging" %% "scala-logging" % "3.9.2"
```
pomadchin: Let's use [log4s](https://github.com/Log4s/log4s) since geotrellis uses it and it looks like we don't need scala-logging anymore.

PR #209, opened 2020-02-18
Fix interpolation w/ duplicate values

### LINKED ISSUE #111: Color interpolation sometimes fails
stack trace:
```
java.lang.IllegalArgumentException: step cannot be 0.
        at scala.collection.immutable.NumericRange$.count(NumericRange.scala:279)
        at scala.collection.immutable.NumericRange.numRangeElements$lzycompute(NumericRange.scala:52)
        at scala.collection.immutable.NumericRange.numRangeElements(NumericRange.scala:51)
        at scala.collection.immutable.NumericRange.length(NumericRange.scala:54)
        at scala.collection.immutable.NumericRange.foreach(NumericRange.scala:72)
        at scala.collection.generic.Growable$class.$plus$plus$eq(Growable.scala:59)
        at scala.collection.mutable.ListBuffer.$plus$plus$eq(ListBuffer.scala:183)
        at geotrellis.server.ogc.Render$.linearInterpolationBreaks(Render.scala:40)
        at geotrellis.server.ogc.ColorRampStyle.renderImage(OgcStyle.scala:64)
        at geotrellis.server.ogc.Render$.apply(Render.scala:16)
        at geotrellis.server.ogc.wms.WmsView$$anonfun$responseFor$1$$anonfun$apply$5.apply(WmsView.scala:91)
        at geotrellis.server.ogc.wms.WmsView$$anonfun$responseFor$1$$anonfun$apply$5.apply(WmsView.scala:89)
        at cats.effect.internals.IORunLoop$.cats$effect$internals$IORunLoop$$loop(IORunLoop.scala:139)
        at cats.effect.internals.IORunLoop$RestartCallback.signal(IORunLoop.scala:351)
        at cats.effect.internals.IORunLoop$RestartCallback.run(IORunLoop.scala:362)
        at cats.effect.internals.Trampoline.cats$effect$internals$Trampoline$$immediateLoop(Trampoline.scala:70)
        at cats.effect.internals.Trampoline.startLoop(Trampoline.scala:36)
        at 
```

with conf:
```
	styles = [
            {
                name = "greyscale2"
                title = "Greyscale Ramp"
                type = "colorrampconf"
                colors = ${color-ramps.slope}
                min-render = 0
                max-render = 80
                stops = 255
            }
        ]
		
	"slope": [
        0x000000FF, 0xFFFFFFFF
    ] 
```

moradology:
## Overview

Interpolation can sometimes attempt to construct breaks from duplicate values. If we don't check for this, a step size of 0 can be used. This leads to an exception which we can avoid if we simply construct an empty list when the desired step size is 0 (which makes sense because there's nothing to interpolate)

### Checklist

## Testing Instructions

Run `ogc-example` project's `RenderSpec`

Closes #111

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated MAML up to 0.6.0 [#199](https://github.com/geotrellis/geotrellis-server/pull/199)
 
 ### Fixed
+- Fixed color interpolation bug related to constructing a range when the step is 0 [#111](https://github.com/geotrellis/geotrellis-server/issues/111)
 - Non-integer (floating point) ColorMap keys now work with or without being quoted [#187](https://github.com/geotrellis/geotrellis-server/issues/187)
 - Missing `<ows:Title>` and `<ows:Abstract>` elements in WCS GetCapabilities response [#114](https://github.com/geotrellis/geotrellis-server/issues/114) 
 - Layer definition elements unused in WMS GetCapabilities response [#115](https://github.com/geotrellis/geotrellis-server/issues/115)
--- a/ogc-example/src/test/scala/geotrellis/server/ogc/RenderSpec.scala
+++ b/ogc-example/src/test/scala/geotrellis/server/ogc/RenderSpec.scala
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2020 Azavea
+ *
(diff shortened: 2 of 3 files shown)


pomadchin:
An extra question, do failure configuration `https://github.com/geotrellis/geotrellis-server/issues/111#issue-444005272` work after applying the fix?

moradology:
This fix should work for that case! The actual source of the pains is from clamping the breaks: because duplication will likely be introduced during clamping, we need to handle the case where there's no step from the current value to the next

PR #213, opened 2020-02-24
Use linspace to interpolate breaks

### LINKED ISSUE #190: Min-render and Max-render don't apply properly on colorrampconf
In the style section of the configuration file, when we add the min-render and max-render parameters for color-ramp, they do not apply correctly. The color-ramp uses the min and max pixel values from the raster layer to define the ramp for each tile.

For example, in the image below, the min-render and max-render parameters are not taken into account for rendering. There is a break between the north tile and the south tile.
 
image

image

moradology:
## Overview

As of now, the `ColorRamp` based rendering strategy in GT-Server suffers from a couple of related issues. 1. It is apparent from tiled results that calculations *per tile* are happening (which is not desirable if we have a heuristic histogram per band from which min/max can be derived) 2. The interpolation that's happening isn't truly linear

This PR uses the GT utility `linspace` to ensure the correct interpolated breaks are being constructed and simplifies the `ColorRamp` rendering process

### Checklist

## Testing Instructions

- Run `OgcStyleSpec`

Closes #190


ogc/src/main/scala/geotrellis/server/ogc/OgcStyle.scala
```diff
@@ -59,34 +60,45 @@ case class ColorRampStyle(
...
+        logger.warn(s"No minimum render value found, using ${Double.MinValue}")
+        Double.MinValue
+      }
+    val maxBreak = maxRender
+      .orElse(hists.head.maxValue())
+      .getOrElse {
+        logger.warn(s"No maximum render value found, using ${Double.MaxValue}")
+        Double.MaxValue
+      }
+    linspace(minBreak, maxBreak, numBreaks)
```
pomadchin: :+1:

moradology:
image

PR #214, opened 2020-02-24
Logging is currently a bit chatty

moradology:
## Overview

Logging is in the process of being refined and is currently far too noisy to be of general utility. This quiets it down a bit and provides more options for environment variable tuning

### DIFF
--- a/ogc-example/src/main/resources/application.conf
+++ b/ogc-example/src/main/resources/application.conf
@@ -256,7 +256,15 @@ layers = {
             zoom = 14
             band-count = 1
         }
-        styles = []
+        styles = [
+            {
+                type = "colorrampconf"
+                name = "Elevation Ramp"
+                title = "Elevation Ramp"
+                colors = ${color-ramps.elevation}
+                stops = 90
+            }
+        ]
     }
     addition-house-income = {
         type = "mapalgebrasourceconf"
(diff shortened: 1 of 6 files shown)


ogc-example/src/main/resources/application.conf
```diff
@@ -256,7 +256,15 @@ layers = {
             zoom = 14
             band-count = 1
         }
-        styles = []
+        styles = [
```
moradology: Not a logging change, but another quality of life improvement

ogc-example/src/main/resources/logback.xml
```diff
@@ -5,7 +5,9 @@
     </encoder>
   </appender>
 
-  <logger name="geotrellis.server.ogc" level="${GT_OGC_LOG_LEVEL:-INFO}"/>
+  <logger name="geotrellis.store" level="${GT_STORE_LOG_LEVEL:-WARN}"/>
+  <logger name="geotrellis.server.ogc" level="${GT_OGC_LOG_LEVEL:-DEBUG}"/>
+  <logger name="org.http4s" level="${HTTP4S_LOG_LEVEL:-WARN}"/>
```
pomadchin: Can we include new params descritpion into https://github.com/geotrellis/geotrellis-server/blob/develop/ogc-example/docs/usage.md?

PR #216, opened 2020-02-24
Fix/provide appropriate warnings for colormap configuration

moradology:
## Overview

A prior fix for HOCON's not properly handling `Double` keyed maps could sometimes run into trouble if key values were not padded to the tenths place (0 must be 0.0). Additionally, values could be thrown out. Values are no longer thrown out and the need for padding or else quoting has been called out in documentation

### DIFF
--- a/ogc-example/src/main/scala/geotrellis/server/ogc/conf/package.scala
+++ b/ogc-example/src/main/scala/geotrellis/server/ogc/conf/package.scala
@@ -84,12 +84,11 @@ package object conf {
         v.valueType match {
           case ConfigValueType.OBJECT =>
             val confmap = v.asInstanceOf[ConfigObject].asScala
-            val ans = confmap.map { case (ck, cv) =>
+            confmap.map { case (ck, cv) =>
               val key = k + "." + ck
               val value = cv.unwrapped.asInstanceOf[String]
               key -> value
             }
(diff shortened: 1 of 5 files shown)


ogc-example/src/main/scala/geotrellis/server/ogc/conf/package.scala
```diff
@@ -73,20 +73,26 @@ package object conf {
...
    *  strings specifically when we know we're trying to decode a ColorMap.
+   * 
+   * @note It is currently difficult to handle double-keyed maps. A workaround
+   * has been provided, but it only works with doubles that explicitly decimal
+   * pad to tenths (0.0 is OK, 0 is to be avoided)
    */
   implicit def colormapReader: ConfigReader[ColorMap] =
     ConfigReader[Map[String, ConfigValue]].map { cmap =>
-      val numericMap = cmap.map({ case (k, v) =>
+      val numericMap = cmap.flatMap({ case (k, v) =>
```
pomadchin: Btw a code style question, do we need `({` here?

pomadchin: Also, flatMaping / maping over a map produces a map; so you don't need `.toMap` conversions

PR #217, opened 2020-02-24
Add ability to configure subset of OGC services

### LINKED ISSUE #151: Cannot configure only one type of web map service
In the configuration file, when I try to configure only two of three web map service, the server cannot start. For example, if I configure WMS and WCS, but not WMST, the server cannot start.
I have the same problem if i configure all of types of web map service but that I keep "layer-definitions" list empty.

 Is this possible to configure only type of web map service i need ?

moradology:
## Overview

It shouldn't be necessary to run a WMS service in order to run a WCS service. This PR makes all top level service configurations optional; empty configuration blocks are tolerated

All supplied:
image

Only WMTS supplied:
image

Closes #151

### DIFF
--- a/ogc-example/src/main/scala/geotrellis/server/ogc/Main.scala
+++ b/ogc-example/src/main/scala/geotrellis/server/ogc/Main.scala
@@ -97,6 +97,9 @@ object Main extends CommandApp(
           CORS(routes)
         }
 
+        def logOptState[A](opt: Option[A], upLog: String, downLog: String): Unit =
+          opt.fold(logger.info(downLog))({ _ => logger.info(upLog) })
+
         val stream: Stream[IO, ExitCode] = {
           import Conf._
           for {
@@ -107,11 +110,11 @@ object Main extends CommandApp(
               .collect { case ssc @ SimpleSourceConf(_, _, _, _) => ssc.models }
               .toList
               .flatten
-            _ <- Stream.eval(IO { conf.wms.fold(
-              logger.info(ansi"%red{WMS configuration detected}")
-            )({ _ =>
-              logger.info(ansi"%green{WMS configuration detected}, starting Web Map Service")
-            })})
+            _ <- Stream.eval(IO(logOptState(
+              conf.wms,
+              ansi"%green{WMS configuration detected}, starting Web Map Service",
+              ansi"%red{No WMS configuration detected}, unable to start Web Map Service"
+            )))
             wmsModel = conf.wms.map { svc =>
               WmsModel(
                 svc.serviceMetadata,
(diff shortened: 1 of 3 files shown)


ogc-example/src/main/scala/geotrellis/server/ogc/Main.scala
```diff
@@ -106,20 +106,44 @@ object Main extends CommandApp(
...
-            wmtsModel = WmtsModel(
-              conf.wmts.serviceMetadata,
-              conf.wmts.tileMatrixSets,
-              conf.wmts.layerSources(simpleSources)
-            )
-            wcsModel = WcsModel(
-              conf.wcs.serviceMetadata,
-              conf.wcs.layerSources(simpleSources)
-            )
+            _ <- Stream.eval(IO {
```
CloudNiner: I was confused for a second as to why this block was here, but its just so that we get logger output on both cases. 

Since each of these blocks are coupled to the ones immediately following, would a `conf.wms.fold()()` work for each of these? In the default case you can log and pass through None, in the valid case you could add the logger statement to the existing map block.

I don't feel strongly about it, just thinking of ways to try to merge these two blocks in a sensible way.

pomadchin: :+1: to using `conf.wms.fold` instead of `conf.wms.isDefined`

ogc-example/src/main/scala/geotrellis/server/ogc/Main.scala
```diff
@@ -106,38 +106,35 @@ object Main extends CommandApp(
               .values
               .collect { case ssc@SimpleSourceConf(_, _, _, _) => ssc.model }
               .toList
-            _ <- Stream.eval(IO {
-              if (conf.wms.isDefined)
-                logger.info(ansi"%green{WMS configuration detected}, starting Web Map Service")
-              else
-                logger.info(ansi"%red{WMS configuration detected}")
-            })
+            _ <- Stream.eval(IO { conf.wms.fold(
```
pomadchin: yea, the syntax is `option.fold(item)(value => smth)`

ogc-example/src/main/scala/geotrellis/server/ogc/Main.scala
moradology: I believe this change makes things less readable 🤷‍♂

ogc-example/src/main/scala/geotrellis/server/ogc/Main.scala
```diff
@@ -107,20 +107,41 @@ object Main extends CommandApp(
...
+              logger.info(ansi"%red{No WMTS configuration detected}")
+            })})
+            wmtsModel = conf.wmts.map { svc =>
+              WmtsModel(
+                svc.serviceMetadata,
+                svc.tileMatrixSets,
+                svc.layerSources(simpleSources)
+              )
+            }
+            _ <- Stream.eval(IO { conf.wcs.fold(
```
pomadchin: @moradology mb it's possible to move it into some separate logging function?

```scala
def serviceLogging(serviceName: String) =
  Stream.eval( 
    IO { conf.wmts.fold(
      logger.info(ansi"%green{WMTS configuration detected}, starting Web Map Tiling Service")
    ){ _ => logger.info(ansi"%red{No WMTS configuration detected}") } }
  )
```

ogc-example/src/main/scala/geotrellis/server/ogc/Main.scala
CloudNiner: Bah. Yeah, I agree with you. Sorry about the poor suggestion -- it was more of musing level of comment.

pomadchin:
Changelog added here https://github.com/geotrellis/geotrellis-server/commit/11ac302731e07fa23ca13ec8f2a2b7d41460c51e

PR #222, opened 2020-02-26
Load RasterSources via SPI

### LINKED ISSUE #174: Use RasterSourceProvider to read COGs
Currently we allow defining a COG layer through [`RasterSourceConf`](https://github.com/geotrellis/geotrellis-server/blob/develop/ogc-example/src/main/scala/geotrellis/server/ogc/conf/RasterSourceConf.scala) and supports COG reads through `GeoTiffRasterSource`. However, there is no way to specify using `GDALRasterSource`.

There configuration for the project should use new  [`RasterSourceProvider`](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/RasterSourceProvider.scala)

This would switch the source configuration to be uniform URI that could be interpreted by the provider interface.

Ref:
- https://github.com/locationtech/geotrellis/blob/master/store/src/main/scala/geotrellis/store/GeoTrellisRasterSourceProvider.scala
- https://github.com/locationtech/geotrellis/blob/master/gdal/src/main/scala/geotrellis/raster/gdal/GDALRasterSourceProvider.scala
- https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/geotiff/GeoTiffRasterSourceProvider.scala

CloudNiner:
## Overview

This PR switches GT Server to use the RasterSource Service Provider Interface, allowing our server configuration to support any type of RasterSource that can be loaded by SPI, not just GeoTrellisRasterSource. The `sources` block in application.conf is now just a list of valid RasterSource URIs.

Unfortunately I had to add a provider for GeoTrellisRasterSourceLegacy, and it looks like the SPI loads that one first so it matches for GT layers. If we're not confident that the SPI will always load the legacy provider first in the iterated list, I can add a new load method that filters out the current GeoTrellisRasterSource. We'd just have to revert to the standard RasterSource apply method when we remove the legacy code.

This PR depends on a new snapshot release of GT, because without it, GeoTrellisRasterSources won't load via SPI due to some bugs in `geotrellis.store.GeoTrellisPath`: https://github.com/locationtech/geotrellis/pull/3191.

The `sources` block is now just a list of valid RasterSource URIs that can be loaded via SPI.

### Checklist

### Demo

Screen Shot 2020-02-26 at 10 32 35 AM

## Testing Instructions

- The example application.conf should work
- It should be possible to construct and load a new layer that references GeoTiff or GDAL raster sources (not yet tested by me)

Closes #174


pomadchin:
@CloudNiner could you update the change log?

CloudNiner:
Also just tested loading GeoTiffRasterSource and GDALRasterSource with this tiff: `s3://azavea-datahub/HRSL/v1/pop/hrsl_pri_pop.tif`.  Noting that use of GDALRasterSource was waaaay faster than GeoTiffRasterSource

Here's the result in GT server when force loaded with the `gdal+` prefix, e.g.:
```
layers = {
    hrsl-pri {
      type = "simplesourceconf"
      name = "HRSL PRI"
      title = ""
      sources = [
        "gdal+s3://azavea-datahub/HRSL/v1/pop/hrsl_pri_pop.tif"
      ]
      styles = []
    }
}
```
Screen Shot 2020-02-26 at 11 35 52 AM

CloudNiner:
Tested working with the GT 3.2.1 snapshot and PR is updated and rebased. Going to merge once CI is green.

pomadchin:
💯

Issue #219, opened 2020-02-25
GeoTrellisRasterSource temporal build improvements


pomadchin:
At this point [GeoTrellisRasterSource.build function](https://github.com/geotrellis/geotrellis-server/blob/571ea30483b72f643acbbea031e66071cd9680e9/core/src/main/scala/geotrellis/store/GeoTrellisRasterSourceLegacy.scala#L270-L299) creates a set of GeoTrellisRasterSources from a temporal layer, making an assumption that for each time bin there is a single spatial slice. This is not true in the most cases in real world:

1. Instead of an optional temporal resolution we need to pass a list of dates that we know are present in the layer,
2. Have an optional ability to read the list of dates that correspond to each temporal slice from the AttributeStore (this can be done as some custom attribute saved into the AttributeStore).
3. Creation of each new GeoTrellisRasterSource for each date is expensive now, since for each new instance we read all the metadata everytime for each slice, though it is the same across all of them.

echeipesh:
Fresh after reviewing the Query DSL PR these are my notes on what the work probably should be, take it for FWIW:

- We need to apply DSL on something other than `List[RasterSource]` because premature creation of `RasterSource` instances is potentially expensive.
- We can use `List[OgcSourceConf]` since that's what we're reading out anyway.
- `OgcSourceConf` should be extended to provide list of dates for the file.
- For `OgcSourceConf` based on GT RasterSource these should be read from `times` attribute in the AttributeStore for zoom level 0.
- We should expect time in ISO 8601 format

CloudNiner:
I'm releasing my assignment on this as it wasn't strictly necessary in the implementation of #223. Going to move onto other stuff first and pull this again if it becomes a bottleneck.

CloudNiner:
A small SpaceTimeKey layer using landsat 8 data is available at s3://geotrellis-test-non-public/spacetimekey-test, layer name `spacetimekey-test`, zoom `14`. I generated it [with the code here](https://gist.github.com/CloudNiner/adf1c4bd07aa4043b17b5fc9ccf3667e) as part of #157.

CloudNiner:
Copying these "where to start" notes from a conversation with Grisha just now:

[...]

PR #223, opened 2020-02-27
First pass DescribeCoverage with TemporalDomain support

### LINKED ISSUE #211: WCS DescribeCoverage may include time TemporalDomain
Along with a [SpatialDomain](https://github.com/geotrellis/geotrellis-server/blob/develop/ogc/src/main/scala/geotrellis/server/ogc/wcs/CoverageView.scala#L78) GetCoverage may optionally include information about time sequences supported by the WCS Coverage. This is the task to modify GetCoverage to return an appropriate temporal information.

CloudNiner:
## Overview

Adds temporal DescribeCoverages support for WCS layers. 

As yet untested, looking for comments on the general strategy taken here so far. Planning to test on a GT layer with 365 temporal slices, I expect to need to address #219 for that to be reasonably performant. For now, though, storing the time value for each OGCSource should be relatively fine as its a lazy val.

### Demo

I was able to perform a DecscribeCoverages request after loading up an old SpaceTimeKey layer for LOCA climate data for the year 2020 that was generated awhile back. Using the configuration:
```
loca-rcp85 {
        type = "simplesourceconf"
        name = "loca-rcp85"
        title = "LOCA RCP8.5 2020-2021"
        source = {
            type = "geotrellis"
            catalog-uri = "s3://development-us-east-1-climate-data/spacetimekey-2020-test"
            layer = "loca-rcp85"
            zoom = 0
            band-count = 1
        }
        styles = [
             {
                 name = "red-to-blue"
                 title = "Red To Blue"
                 type = "colorrampconf"
                 colors = ${color-ramps.red-to-blue}
                 stops = 64
             }
        ]
    }
```
and appropriate s3 credentials for that account, I got a DescribeCoverages that looks like this:

Screen Shot 2020-02-27 at 2 00 22 PM

🔥 

This matches the schema presented in `opengis/src/main/xsd/wcs/1.1.1/Examples/exampleCoverageDescription2.xml`

### Notes

Optional. Ancillary topics, caveats, alternative strategies that didn't work out, anything else.

## Testing Instructions

Closes #211


ogc/src/main/scala/geotrellis/server/ogc/OgcSourceRepository.scala
```diff
@@ -22,23 +22,23 @@ import io.circe.Json
...
-
 case class OgcSourceRepository(store: List[OgcSource]) extends Repository[List] {
   def find(query: Query): List[OgcSource] = OgcSourceRepository.eval(query)(store)
 }
 
 object OgcSourceRepository {
   import geotrellis.store.query.QueryF._
 
-  def algebgra: Algebra[QueryF, List[OgcSource] => List[OgcSource]] = Algebra {
+  def algebra: Algebra[QueryF, List[OgcSource] => List[OgcSource]] = Algebra {
```
pomadchin: nice catch

ogc/src/main/scala/geotrellis/server/ogc/OgcSourceRepository.scala
```diff
@@ -22,23 +22,23 @@ import io.circe.Json
...
     case Nothing()           => _ => Nil
     case All()               => identity
     case WithName(name)      => _.filter(_.name == name)
     case WithNames(names)    => _.filter(rs => names.contains(rs.name))
-    case At(t, fn)           => _.filter(_.metadata.attributes.get(fn).map(ZonedDateTime.parse).fold(false)(_ == t))
-    case Between(t1, t2, fn) => _.filter {
-      _.metadata.attributes.get(fn).map(ZonedDateTime.parse).fold(false) { current => t1 >= current && t2 < current }
+    case At(t, _)           => _.filter(_.time.getOrElse(null) == t)
```
pomadchin: `null`?

ogc/src/main/scala/geotrellis/server/ogc/wcs/CoverageView.scala
```diff
@@ -68,6 +75,17 @@ object CoverageView {
     val ex = re.extent
     val llex = llre.extent
     val Dimensions(w, h) = re.dimensions
+    val temporalInstants: Seq[DataRecord[String]] = sources.flatMap { s =>
+      // TODO: Attempt to use `GmlDataRecord(TimePositionType(t.toWcsIsoString))`
+      //       fails with a "Missing element label" exception, not sure how this
+      //       doesn't have a CanWriteXML but all the other types do?
```
pomadchin: Is it a question you have?

pomadchin: Ah right, if it is a question, this is a weird thing about the protocol generated. 

Let me see what we can do here, I parsed the question.

ogc/src/main/scala/geotrellis/server/ogc/OgcSourceRepository.scala
CloudNiner: The idea here was to succinctly return false if our OGCSource Option[ZonedDateTime] is None and it looks like `zonedDateTime == null` returns false. Is there a better way to do this comparison? We want None == t to be false and otherwise compare on the two date times.

ogc/src/main/scala/geotrellis/server/ogc/wcs/CoverageView.scala
CloudNiner: Sorta yes sorta no. It's a note to myself to resolve later, but if you have any suggestions in the meantime I'm all ears.

ogc/src/main/scala/geotrellis/server/ogc/OgcSourceRepository.scala
pomadchin: @CloudNiner ` _.filter(_.time.exists(t))`

ogc/src/main/scala/geotrellis/server/ogc/wcs/CoverageView.scala
```diff
@@ -50,7 +55,9 @@ class CoverageView(
 }
 
 object CoverageView {
-  def sourceDescription(source: OgcSource): CoverageDescriptionType = {
+
+  def sourceDescription(sources: List[OgcSource]): CoverageDescriptionType = {
+    val source = sources.head
```
CloudNiner: Just calling out here that we're now explicitly using the first source in the list of valid sources for a given identifier to construct all of the spatial domain information for a layer. This was happening implicitly before because there was a 1 to 1 mapping of identifier to ogc source. With the addition of temporal slices, that's no longer the case. This assumption will continue to work as long as our list of sources all have the same spatial domain and CRS.

ogc/src/main/scala/geotrellis/server/ogc/wcs/CoverageView.scala
pomadchin: @CloudNiner yep, it should always be the 

`TimeSequenceType(GmlDataRecord("timePosition", t.toWcsIsoString) :: GmlDataRecord("timePosition", t2.toWcsIsoString))` 

There is nothing to do with it; I think you're safe to remove this comment. 

I think types here are used for type safety only, but the key / name of the xml record lacks from the generated case classes. It leads to a neccesity of passing them manually. That's what I had to do with all other records as well.

CloudNiner: Ohhhh I see. I missed that its the `OwsDataRecord.apply(value)` that's constrcuting the element name from the passed class tag. 🎉

ogc/src/main/scala/geotrellis/server/ogc/wcs/package.scala
```diff
@@ -45,4 +49,27 @@ package object wcs {
...
+   * ― hh:mm:ss.sss is the time (a two-digit hour and minute, and fractional seconds);
+   * ― Z represents the Coordinated Universal Time (UTC or ―zulu‖) time zone.
+   *
+   * This was excerpted from "WCS Implementation Standard 1.1" available at:
+   * https://portal.ogc.org/files/07-067r5
+   *
+   * @param zdt
+   */
+  implicit class ZonedDateTimeWcsOps(val zdt: ZonedDateTime) {
+    val formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.'000Z'")
```
pomadchin: I think in Java8 ISO formatter is inbuilt: 

```
scala> ZonedDateTime.now().toInstant.toString
res4: String = 2020-02-27T17:33:50.149Z
```

ogc/src/main/scala/geotrellis/server/ogc/wcs/package.scala
```diff
@@ -45,4 +49,27 @@ package object wcs {
...
+    def toWcsIsoString: String = zdt.withZoneSameInstant(ZoneId.of("UTC")).format(formatter)
+  }
```
pomadchin: Btw, instead of `ZoneId.of("UTC")` you can use `ZoneOffset.UTC`

ogc/src/main/scala/geotrellis/server/ogc/wcs/CoverageView.scala
```diff
@@ -68,6 +75,14 @@ object CoverageView {
     val ex = re.extent
     val llex = llre.extent
     val Dimensions(w, h) = re.dimensions
+    val temporalInstants: Seq[DataRecord[String]] = sources.flatMap { s =>
+      s.time.map(t => GmlDataRecord(TimePositionType(t.toWcsIsoString)))
+    }
+    val temporalDomain: Option[TimeSequenceType] = if (temporalInstants.length > 0) {
```
pomadchin: `temporalInstants.nonEmpty`?

ogc/src/main/scala/geotrellis/server/ogc/wcs/package.scala
CloudNiner: Cool, I'll switch to that then

CloudNiner:
@pomadchin ready for another look. Tests pass and I was able to generate the describecoverages request on a real layer. See PR description for proof.

pomadchin:
Oops we forgot about the CHANGELOG here

pomadchin:
Added it here https://github.com/geotrellis/geotrellis-server/commit/1d9c767b001e2021ace9aaa95e923ef7ba5919ee

pomadchin:
** A DISCLAMER for those who are reading through this issue:**
Before merge I added cleanup changes via d3f93e8 without creating a discussion about it that could cause some misunderstandings; In short ~ `toWcsIsoString` was removed since it is _probably not required_, since it can be replaced via `zdt.toInstant.toString` this is arugable. However we can add  smth like `.toISOString` func call that would be valid for both WCS and WMS (if we really want to).

pomadchin:
**A DISCLAMER for those who are reading through this issue**

Before merge I added cleanup changes via d3f93e8 without creating a discussion about it that could cause some misunderstandings; In short ~ `toWcsIsoString` was removed since it is _probably not required_, since it can be replaced via `zdt.toInstant.toString` (this is arugable though). However we can add  smth like `.toISOString` func call that would be valid for both WCS and WMS (if we really want to). 

Sorry for merging it in without any discussion ):. It is a minor thing but it definitely should have happened.

PR #224, opened 2020-02-27
Add interpolated color map styling

### LINKED ISSUE #161: Add interpolation to color configuration for WMS/WMTS
`ColorMap` associates a range of breaks (raster values) with a color.

Currently rendering a raster results snapping each pixel value to the closest break value and assign it corresponding color value. As such the output raster may only have colors explicitly specified in the color ramp. This is how `Discrete` color rendering works in QGIS.

This issue is to add a rendering option that will interpolate option to rendering that takes place in WMS and WMTS endpoints. When enabled the assigned color is going to be interpolated (linearly) in the color space based on distance of the pixel value between two closest break values in the `ColorMap`.

moradology:
## Overview

GT Server already has a `ColorRamp` and `ColorMap` styling but currently lacks a way to define fine grained interpolation between preset poles and their associated colors. This PR rectifies that situation by introducing the `InterpolatedColorMap` and the mechanisms necessary to define/configure them

### Checklist

### Demo

This style conf:
```javascript
    "income": {
        "poles": {
            "0.0": 0xFF0000FF,
            "75000.0": 0x777700FF
            "200000.0": 0x00FF00FF,
        }
        "clip-definition": "clip-both"
    }
```
applied to a median household income layer produces this layer:
image

### Notes

The same parsing difficulties that HOCON creates for regular `ColorMap`s continue to be potential issues here

Closes #161

### DIFF
--- a/ogc-example/docs/conf.md
+++ b/ogc-example/docs/conf.md
@@ -137,8 +137,8 @@ currently supported:
 
 Note above that each layer is configured with 0 or more style
 objects. Each provided style object defines an alternative coloring
-strategy. At the highest level, this server has two flavors of styling
-object: `ColorRamp` and `ColorMap` based.
+strategy. At the highest level, this server has three flavors of styling
+object: `ColorRamp`, `ColorMap` and `InterpolatedColorMap` based.
 
 Color ramps are just a list of colors and it is up to the server to
 produce the statistics necessary to make decisions about where (in
@@ -170,22 +170,6 @@ color-ramps = {
     ]
 ```
 
-A word of caution about color ramps: statistically derived rendering
-strategies can mislead if the entire layer being colored is markedly
-different from a well known distribution. If, for instance, we have NDVI
-values from a recent wildfire and our layer contains little information
-outside of the damaged area, statistically derived styling will see a
-distribution skewing towards -1 (or 0, depending on how the NDVI is
-calculated) and is therefore likely to use too many color breaks in the
-low end of the NDVI range.
-
-In a cases like the one just described, you're better off using a color map
-that is defined specifically for the range of values capable of being
-represented (e.g. for NDVI -1 is red, 0 is yellow, and 1 is green
-regardless of the distribution of NDVI values I happen to be looking
-at). If only the minimum and maximum values are known, they may
-optionally be specified in color ramp configuration (as above).
-
 A color map style definition from `application.conf` (note the HOCON
 reference):
 ```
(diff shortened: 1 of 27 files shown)


ogc-example/docs/conf.md
```diff
@@ -170,22 +170,6 @@ color-ramps = {
...
-calculated) and is therefore likely to use too many color breaks in the
-low end of the NDVI range.
-
-In a cases like the one just described, you're better off using a color map
-that is defined specifically for the range of values capable of being
-represented (e.g. for NDVI -1 is red, 0 is yellow, and 1 is green
-regardless of the distribution of NDVI values I happen to be looking
-at). If only the minimum and maximum values are known, they may
-optionally be specified in color ramp configuration (as above).
-
```
moradology: No longer relevant after the fix to `ColorRamp` behavior (we don't use quantile breaks any longer)

ogc-example/src/main/scala/geotrellis/server/ogc/conf/Conf.scala
```diff
@@ -42,6 +43,7 @@ case class Conf(
 )
 
 object Conf {
+  implicitly[ConfigReader[ClipDefinition]]
```
pomadchin: Is it neccesary here? Looks like some test code that got into the PR

ogc/src/main/scala/geotrellis/server/ogc/style/InterpolatedColorMap.scala
```diff
@@ -0,0 +1,93 @@
...
+package geotrellis.server.ogc.style
+
+import geotrellis.raster.render._
+import geotrellis.raster.{RGBA => _, _}
+
+import scala.math._
+import java.util.Arrays.binarySearch
+
+
+case class InterpolatedColorMap(
```
pomadchin: Should this be a part of GeoTrellis?

ogc/src/main/scala/geotrellis/server/ogc/style/LegendModel.scala
```diff
@@ -0,0 +1,8 @@
+package geotrellis.server.ogc.style
+
+case class LegendModel(
```
pomadchin: You can run `headerCreate` and `test:headerCreate` to generate headers, it won't regenerate headers for the files that already have it.

ogc/src/main/scala/geotrellis/server/ogc/style/InterpolatedColorMap.scala
moradology: I think that's a great idea but one which should be done as part of a larger push to handle common rendering strategies; do you have some thoughts about how best to begin working these changes in?

pomadchin: @moradology at least to create an issue probably to think about it :+1: and you're probably the person who knows better how to describe it / how to bring it into GT. But it in general looked for me like a functionality that can be required in other projects as well (like RF).

Issue #225, opened 2020-03-02
Support XML responses for WCS GetCoverage


CloudNiner:
The WCS 1.1.1 spec supports returning a GetCoverage response as an XML document according to https://portal.opengeospatial.org/files/07-067r2

We currently implement:
```
10.3.1 Introduction
The normal response to a valid GetCoverage operation request shall be a single coverage extracted from the coverage requested, with the specified spatial reference system, bounding box, size, format, and range subset.
```

But we could also support:
```
10.3.11 Response encodings
10.3.11.1 Overview
The encoding of the GetCoverage response consists of a Coverages XML document (see 10.3.11.2 below), packaged and bundled according to the request encoding and the value of the ―store‖ parameter (see Table 24). WCS servers shall implement response encodings according to Table 31.
```

This additional feature would allow us to support much more complete TimeSequence queries for arbitrary lists of TimePeriods or TimeIntervals.

It's likely that this implementation would also require implementing the `store` parameter to fully support the spec.

To verify that these types of requests are not currently supported by gt server, I made a request to:

http://localhost:9000/?SERVICE=WCS&VERSION=1.1.1&REQUEST=GetCoverage&FORMAT=xml&IDENTIFIER=us-ned&<...>

And got:

`Error parsing parameters: Unsupported format: 'xml'`.

CloudNiner:
Noting here that it doesn't look like GeoServer supports temporal queries at all, their WCS Request Builder demo doesn't allow you to choose a time slice:
https://docs.geoserver.org/stable/en/user/services/wcs/requestbuilder.html#wcs-request-builder

QGIS supports the TimeSequence parameter, at least it [exists in the code](https://github.com/qgis/QGIS/blob/final-3_10_2/src/providers/wcs/qgswcsprovider.cpp#L748). Further testing in #157 might reveal further details.

PR #226, opened 2020-03-02
Add timesequence param for WCS GetCoverage requests

### LINKED ISSUE #157: Add TimeSequence parameter support to GetCoverage request
When querying layers that have temporal component, for instance GeoTrellis spatiotemporal layer or a STAC catalog it should be possible to specify the time range for the query.

Screen Shot 2019-10-09 at 3 42 20 PM

The parameter is described further in OGC specification document for WCS 1.1.1: https://portal.opengeospatial.org/files/07-067r2

This issue covers parsing the query, routing it to the catalog model and return an actual raster response.

It is possible that `TimeSequence` parameter will be specified as a time range or a sequence of dates.Because we expect to be querying data-cube of rasters that already exist the expected result is a coverage that falls on or with-in times described. This parameter is specific to version 1.1.* of WCS protocol.

This is not officially part of the specification but seems to be a reasonable interpretation for our case.

CloudNiner:
## Overview

This PR adds basic temporal support for querying a single TimeInterval or TimePeriod for a given WCS layer. The spec supports querying over multiple intervals or periods but that's not really compatible with GT server currently returning a single raster for all GetCoverage requests. I opened #225 to address expanded temporal querying in the future.

Before calling this issue complete, I think we should still address the following:
- [x] ~~Generate or find a simple GeoTrellis layer with a temporal component that we can use for testing~~ Generated and a note added to #219. Further work to improve performance will happen there.
- [x] ~~Figure out how to get QGIS to include the temporal component when configured to render WCS layers~~ Looks to be not trivial. New issue created in our private team repo due to client specific concerns since it doesn't look like QGIS doesn't natively support temporal params on WCS layers.

### Checklist

### Demo

I again used the same LOCA RCP 85 climate data layer I used in #223 which has daily slices for one year.

GetCoverage request with a TimePeriod param `timesequence=2020-01-01T00:00:00Z` correctly filters to a single source available in the layer:
Screen Shot 2020-03-02 at 3 43 53 PM

GetCoverage request with a TimeInterval param `TIMESEQUENCE=2020-01-15T00:00:00Z/2020-02-15T00:00:00Z` correctly filters to the 31 slices valid during that range:
Screen Shot 2020-03-02 at 4 17 55 PM

## Testing Instructions

Configure the same layer as in #223 (LOCA RCP85) and issue a GetCoverage request that includes the timesequence parameter such as:

http://localhost:9000/?SERVICE=WCS&VERSION=1.1.1&REQUEST=GetCoverage&FORMAT=image/geotiff&IDENTIFIER=loca-rcp85&TIMESEQUENCE=2020-01-01T00:00:00Z&BOUNDINGBOX=38.63767820773930595%2C-96.15608740894901985%2C38.76232179226069974%2C-95.84391259105098015%2Curn%3Aogc%3Adef%3Acrs%3AEPSG%3A%3A4326&GRIDBASECRS=urn%3Aogc%3Adef%3Acrs%3AEPSG%3A%3A4326&GRIDCS=urn%3Aogc%3Adef%3Acs%3AOGC%3A0.0%3AGrid2dSquareCS&GRIDTYPE=urn%3Aogc%3Adef%3Amethod%3AWCS%3A1.1%3A2dSimpleGrid&GRIDORIGIN=38.76232179226069974%2C-96.15608740894901985&GRIDOFFSETS=-0.06232179226069453%2C0.0624349635796051

This particular layer will timeout attempting to serve the query as it requires pulling a large sum of data across a tiled data source. 😞 I still haven't constructed or found a good test layer that includes a temporal dimension. 

Closes #157

### DIFF
--- a/ogc/src/main/scala/geotrellis/server/ogc/OgcTimeInterval.scala
+++ b/ogc/src/main/scala/geotrellis/server/ogc/OgcTimeInterval.scala
@@ -1,0 +1,49 @@
+package geotrellis.server.ogc
+
+import java.time.ZonedDateTime
+
+/**
+  * Represents the TimeInterval and TimePosition types used in TimeSequence requests
+  *
+  * If end is provided, a TimeInterval is assumed. Otherwise, a TimePosition.
+  *
+  * @param start The start time for this TimePosition or TimeInterval
+  * @param end The end time for this TimeInterval. If None, a TimePosition is assumed from the
+  *              start param.
+  * @param interval ISO 8601:2000 provides a syntax for expressing time periods: the designator P,
+  *                 followed by a number of years Y, months M, days D, a time designator T, number
+  *                 of hours H, minutes M, and seconds S. Unneeded elements may be omitted. Here are
+  *                 a few examples:
+  *                 EXAMPLE 1 -  P1Y, 1 year
+  *                 EXAMPLE 2 - P1M10D, 1 month plus 10 days
+  *                 EXAMPLE 3 - PT2H, 2 hours
+  *
+  *                 @note This param is not validated. It is up to the user to ensure that it is
+  *                       encoded directly
+  */
+final case class OgcTimeInterval(start: ZonedDateTime,
+                                 end: Option[ZonedDateTime],
+                                 interval: Option[String]) {
+  override def toString: String = {
+    end match {
+      case Some(e) => s"${start.toInstant.toString}/${e.toInstant.toString}${interval.map("/" + _).getOrElse("")}"
+      case _ => start.toInstant.toString
+    }
+  }
+}
+
+object OgcTimeInterval {
+  def apply(timePeriod: ZonedDateTime): OgcTimeInterval = OgcTimeInterval(timePeriod, None, None)
+
+  def fromString(timeString: String): OgcTimeInterval = {
+    val timeParts = timeString.split("/")
+    timeParts match {
+      case Array(start, end, interval) =>
+        OgcTimeInterval(ZonedDateTime.parse(start), Some(ZonedDateTime.parse(end)), Some(interval))
+      case Array(start, end) =>
+        OgcTimeInterval(ZonedDateTime.parse(start), Some(ZonedDateTime.parse(end)), None)
+      case Array(start) => OgcTimeInterval(ZonedDateTime.parse(start))
+      case _ => throw new UnsupportedOperationException("Unsupported string format for OgcTimeInterval")
+    }
+  }
+}
--- a/ogc/src/main/scala/geotrellis/server/ogc/wcs/WcsModel.scala
+++ b/ogc/src/main/scala/geotrellis/server/ogc/wcs/WcsModel.scala
@@ -24,14 +24,17 @@ case class WcsModel(
   sources: OgcSourceRepository
 ) {
 
-  def getLayers(p: GetCoverageWcsParams): List[OgcLayer] =
-    sources
-      .find(p.toQuery)
-      .map {
+  private val logger = org.log4s.getLogger
+
+  def getLayers(p: GetCoverageWcsParams): List[OgcLayer] = {
+    val filteredSources = sources.find(p.toQuery)
+    logger.debug(s"Filtering sources: ${sources.store.length} -> ${filteredSources.length}")
+    filteredSources.map {
         case SimpleSource(name, title, source, _, styles) =>
           SimpleOgcLayer(name, title, p.crs, source, None)
         case MapAlgebraSource(name, title, sources, algebra, _, styles) =>
           val simpleLayers = sources.mapValues { rs => SimpleOgcLayer(name, title, p.crs, rs, None) }
           MapAlgebraOgcLayer(name, title, p.crs, simpleLayers, algebra, None)
       }
+  }
 }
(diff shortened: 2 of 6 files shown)


ogc/src/main/scala/geotrellis/server/ogc/OgcTimeInterval.scala
```diff
@@ -0,0 +1,50 @@
...
+  */
+final case class OgcTimeInterval(start: ZonedDateTime,
+                                 end: Option[ZonedDateTime],
+                                 interval: Option[String]) {
+  override def toString: String = {
+    if (end.isDefined) {
+      s"${start.toInstant.toString}/${end.get.toInstant.toString}${interval.map("/" + _).getOrElse("")}"
+    } else {
+      start.toInstant.toString
+    }
```
pomadchin: Can be fold or match. It is not a void call so we can't really use unsafe `option.get` calls.

```scala
end.fold(start.toInstant.toString) { end => 
  s"${start.toInstant.toString}/${end.toInstant.toString}${interval.map("/" + _).getOrElse("")}" 
}
```

or 

```scala
end match {
  case Some(end) => s"${start.toInstant.toString}/${end.toInstant.toString}${interval.map("/" + _).getOrElse("")}"
  case _ => start.toInstant.toString
}
```

ogc/src/main/scala/geotrellis/server/ogc/wcs/WcsParams.scala
```diff
@@ -53,20 +52,37 @@ case class DescribeCoverageWcsParams(version: String, identifiers: Seq[String])
...
+                                 temporalSequence: Option[NonEmptyList[OgcTimeInterval]],
+                                 format: OutputFormat,
+                                 gridBaseCRS: CRS,
+                                 gridCS: URI,
+                                 gridType: URI,
+                                 // GridOrigin is BBOX minx, maxy // swapped in case of a geographic projection
+                                 gridOrigin: (Double, Double),
+                                 // GridOffsets is xres, yres // swapped in case of a geographic projection
+                                 gridOffsets: (Double, Double),
+                                 crs: CRS
```
pomadchin: Do you think such alignment is more readble or you're using a default IDEA formatter?

ogc/src/main/scala/geotrellis/server/ogc/wcs/WcsParams.scala
```diff
@@ -53,20 +52,37 @@ case class DescribeCoverageWcsParams(version: String, identifiers: Seq[String])
...
+    // temporal filter on the first TimeInterval in the list. Revisit when we are
+    // able to utilize all requested TimeIntervals.
+    temporalSequence match {
+      case Some(ts: NonEmptyList[OgcTimeInterval]) =>
+        val timeInterval = ts.head
+        if (timeInterval.end.isDefined) {
+          query and between(timeInterval.start, timeInterval.end.get)
+        } else {
+          query and at(timeInterval.start)
+        }
```
pomadchin: Should be smth like

```scala
timeInterval
  .end
  .fold(query and at(timeInterval.start))(end => query and between(timeInterval.start, end))
```

Or 

```scala
timeInterval.end match {
  case Some(end) => query and between(timeInterval.start, end)
  case _ => query and at(timeInterval.start)
}
```

ogc/src/main/scala/geotrellis/server/ogc/wcs/WcsModel.scala
```diff
@@ -24,14 +24,16 @@ case class WcsModel(
   sources: OgcSourceRepository
 ) {
 
-  def getLayers(p: GetCoverageWcsParams): List[OgcLayer] =
-    sources
-      .find(p.toQuery)
-      .map {
+  def getLayers(p: GetCoverageWcsParams): List[OgcLayer] = {
+    val logger = org.log4s.getLogger
```
pomadchin: logger is usually created per class instance, not per a function call; It's better to put it into WcsModel class private field.

ogc/src/main/scala/geotrellis/server/ogc/wcs/WcsParams.scala
CloudNiner: This was just a formatter choice. Should I revert?

pomadchin: shrug

ogc/src/main/scala/geotrellis/server/ogc/wcs/WcsParams.scala
```diff
@@ -18,12 +18,11 @@ package geotrellis.server.ogc.wcs
 
 import geotrellis.proj4.CRS
 import geotrellis.raster.{CellSize, GridExtent}
-import geotrellis.server.ogc.OutputFormat
+import geotrellis.server.ogc.{OgcTimeInterval, OutputFormat}
 import geotrellis.server.ogc.params.ParamError.UnsupportedFormatError
 import geotrellis.server.ogc.params._
-import geotrellis.store.query._
+import geotrellis.store.query.{at, _}
```
pomadchin: What does that mean? is not `import geotrellis.store.query._` just enough?

CloudNiner:
@pomadchin CHANGELOG added and commit history cleaned up

PR #230, opened 2020-03-12
Add Time Dimension to WMS GetCapabilities

### LINKED ISSUE #185: WMS GetCapabilities may include time Dimensions
For spatial-temporal sources the WMS services should include a `Dimension` element in `GetCapabilities` response that describes the dimensions.

Handling mult-dimensional data is fully discussed in Annex C, Page 51, [WCS 1.3.0 Specification](https://www.opengeospatial.org/standards/wms)

An example of layer description of `time` dimension.
```xml
  <Layer queryable="1">
      <Title>Weather Forecast Data</Title>
      <CRS>CRS:84</CRS> <!-- harmless repetition of common CRS -->

      <EX_GeographicBoundingBox>
        <westBoundLongitude>-180</westBoundLongitude>
        <eastBoundLongitude>180</eastBoundLongitude>
        <southBoundLatitude>-90</southBoundLatitude>
        <northBoundLatitude>90</northBoundLatitude>
      </EX_GeographicBoundingBox>
      <!-- These weather data are available daily from 1999-01-01 through
           2000-08-22. -->
      <Dimension name="time" units="ISO8601" default="2000-08-22">
         1999-01-01/2000-08-22/P1D
      </Dimension>
      <Layer>
        <Name>Clouds</Name>
        <Title>Forecast cloud cover</Title>
      </Layer>
      <Layer>
        <Name>Temperature</Name>
        <Title>Forecast temperature</Title>
      </Layer>
      <Layer>
        <Name>Pressure</Name>
        <Title>Forecast barometric pressure</Title>
             <!-- This Pressure layer is available at several elevations and times. -->
         <Dimension name="elevation" units="EPSG:5030" />
         <Dimension name="time" units="ISO8601" default="2000-08-22">
           1999-01-01/2000-08-22/P1D</Dimension>
         <Dimension name="elevation" units="CRS:88" default="0" nearestValue="1">
           0,1000,3000,5000,10000</Dimension>
      </Layer>
    </Layer>
``` 

This issue assumes that the data model already exists that exposes the temporal information for layers and is only concerned with exposing this information though WMS response.

Connects: https://github.com/geotrellis/geotrellis-server/issues/162

CloudNiner:
## Overview

Not much to add here...

### Checklist

### Demo

Started the server with the newly committed application-spacetimekey.conf and got this back:

Screen Shot 2020-03-12 at 3 27 27 PM

## Notes

WMS, unlike WCS, has a parent layer that defines the ranges for all included child layers. The time range on this parent layer is derived from the min/max times in the child layers. The time interval _can_ have an "interval" component but I'm not generating that here. It would be far too difficult to generate the correct interval from an arbitrary list of individual date times. Interval is an optional part of the time spec anyways.

## Testing Instructions

`./sbt` followed by `project ogcExample` then
```
run --public-url http://127.0.0.1:9000 --interface 0.0.0.0 --port 9000 --conf src/main/resources/application-spacetimekey.conf
```

Can view the result at http://localhost:9000/?SERVICE=WMS&REQUEST=GetCapabilities
or by adding GT server as a new WMS layer in QGis.

Closes #185

### DIFF
--- a/ogc-example/src/main/resources/application-spacetimekey.conf
+++ b/ogc-example/src/main/resources/application-spacetimekey.conf
@@ -1,0 +1,220 @@
+wms = {
+  parent-layer-meta = {
+    name = "Geotrellis WMS Parent Layer"
+    title = "WMS Parent Title"
+    description = "Top level metadata that is inherited by children layers"
+    supported-projections = [
+      4326,
+      3410,
+      3978,
+      4617,
+      3979,
+      3413,
+      26916
+    ]
+  }
+  service-metadata = {
+    name = "WMS"
+    title = "GeoTrellis Service"
+    online-resource = {}
+    keyword-list = {
+      keyword = ["geotrellis", "catalog"]
+    }
+    contact-information = {
+      contact-person-primary = {
+        contact-person = "Eugene Cheipesh"
+        contact-organization = "Azavea"
+      }
+      contact-position = "Developer"
+      contact-address = {
+        address-type = "Office"
+        address = "990 Spring Garden St."
+        city = "Philadelphia"
+        state-or-province = "PA",
+        post-code = "19087",
+        country = "USA"
+      }
+    }
+  }
+  layer-definitions = [
+    ${layers.spacetimetest}
+  ]
+}
+
+wcs = {
+  service-metadata = {
+    identification = {
+      title = "WCS"
+      description = "Geotrellis WCS Service"
+      keywords = []
+      profile = ["http://azavea.com/wcs-profile"]
+      fees = ""
+      access-constraints = []
+    }
+    provider = {
+      name = "Azavea"
+      site = "https://www.azavea.com"
+    }
+  }
+  layer-definitions = [
+    ${layers.spacetimetest}
+  ]
+}
+
+wmts = {
+  service-metadata = {
+    identification = {
+      title = "WMTS"
+      description = "Geotrellis WMTS Service"
+      keywords = []
+      profile = ["http://azavea.com/wmts-profile"]
+      fees = ""
+      access-constraints = []
+    }
+    provider = {
+      name = "Azavea"
+      site = "https://www.azavea.com"
+    }
+  }
+  layer-definitions = []
+  tile-matrix-sets = [
+    {
+      identifier = "GoogleMapsCompatible",
+      supported-crs = 3857,
+      title = "GoogleMapCompatible",
+      abstract = "Google Maps compatible tile matrix set",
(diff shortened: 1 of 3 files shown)


pomadchin:
@CloudNiner does QGIS supports GetCoverage and temporal queries?

ogc/src/main/scala/geotrellis/server/ogc/wms/CapabilitiesView.scala
```diff
@@ -104,6 +115,17 @@ object CapabilitiesView {
...
+  def timeDimension(times: Seq[Option[ZonedDateTime]]): Seq[Dimension] = {
+    val validTimes = times.flatten
+    if (validTimes.isEmpty) {
+      Nil
+    } else if (validTimes.length == 1) {
+      OgcTimeInterval(validTimes.head)
+    } else {
+      OgcTimeInterval(validTimes.min, Some(validTimes.max), None)
+    }
+  }
```
pomadchin: I'm fine with that piece of code

But just to get rid of unsafe `.head` and a length check how about the following:

```scala
validTimes match {
  case Nil         => Nil
  case head :: Nil => OgcTimeInterval(head)
  case list        => OgcTimeInterval(list.min, list.max.some, None) 
}
```

ogc-example/src/main/resources/application-spacetimekey.conf
```diff
@@ -0,0 +1,220 @@
+wms = {
```
pomadchin: Actually this is a pretty cool idea to add this conf.  :100:

ogc/src/main/scala/geotrellis/server/ogc/wms/CapabilitiesView.scala
CloudNiner: TIL that Nil matches an empty Seq:
```scala
scala> Seq() match {
     | case Nil => print("empty")
     | case _ => print("other")
     | }
empty
```

Yeah, this is a much cleaner check. Updated.

ogc-example/src/main/resources/application-spacetimekey.conf
CloudNiner: Yeah. A bit of a caveat. I'm not sure its generated correctly to actually return images in QGis, but it at least is setup so that metadata is correctly generated by gt server. This is also the only temporal layer I've tested our new features against, so I expect bugs/inconsistencies/edge cases as we test against more. We'll likely want to make this example better as we iterate on temporal support.

PR #233, opened 2020-03-19
Expose RasterSource resample methods through configuration options

### LINKED ISSUE #150: Artefacts with hillshade MAML operation
When I use fhillshade MAML operation from the DEM data, It have artefacts on the GetMap result. On the example below, we see vertical and horizontal lines. The request GetMap use projection of original data (EPSG:3857) and a color-maps style.

`
SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&CRS=**EPSG:3857**&BBOX=-8212121.0951458206,10994397.212170906,-8204445.5172946611,11001294.913466308&WIDTH=967&HEIGHT=869&LAYERS=arcticdem2mhillshade&STYLES=greyscalemap&EXCEPTIONS=XML&FORMAT=image/png&BGCOLOR=0xFEFFFF&TRANSPARENT=TRUE&
`
image

We see the same thing for a GetMap request that uses a different projection than the original data.
`
SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&CRS=**EPSG:3978**&BBOX=780813.0364727308,2393130.123206886,786953.49875365547,2398648.2842432079&WIDTH=967&HEIGHT=869&LAYERS=arcticdem2mhillshade&STYLES=greyscalemap&EXCEPTIONS=XML&FORMAT=image/png&BGCOLOR=0xFEFFFF&TRANSPARENT=TRUE
`
image

We see the same thing with color-ramps style.

image

### LINKED ISSUE #229: Configurable ResampleMethod in source definitions
Connects: #150

`SimpleSource` or `MapAlgebraSource` (for each source) should have an option to configure the resample method. Currently all instances default to `NearestNeighbor` resample method which probably cause artifacts in #150

pomadchin:
## Overview

This PR adds resample method as a cofigurable application.conf option

### Checklist

### Demo

bilinear-nn

It's also possible to view it via QGIS & a GeoTrellis server instance:

```bash
$ ./sbt
> project ogcExamples
> run --public-url http://localhost:9000/
```

Select the `us-ned-hillshade` WMS/WMTS/WCS layer!

Closes #229
Closes #150

### DIFF
--- a/core/src/test/scala/geotrellis/server/LayerExtentTest.scala
+++ b/core/src/test/scala/geotrellis/server/LayerExtentTest.scala
@@ -18,17 +18,16 @@ package geotrellis.server
 
 import geotrellis.raster._
 import geotrellis.vector._
-import geotrellis.raster.resample.NearestNeighbor
-
-import org.scalatest._
 
 import scala.concurrent.ExecutionContext
+
+import org.scalatest._
 
 class LayerExtentTest extends FunSuite with Matchers {
   implicit val cs = cats.effect.IO.contextShift(ExecutionContext.global)
 
   test("ability to read a selected extent") {
-    val rt = ResourceTile("8x8.tif", NearestNeighbor)
+    val rt = ResourceTile("8x8.tif")
     val eval = LayerExtent.identity(rt)
     // We'll sample such that the bottom row (from 56 to 64) are excised from the result
     val sampled = eval(Extent(0, 1, 8, 8), CellSize(1, 1)).unsafeRunSync
--- a/core/src/test/scala/geotrellis/server/LayerExtentTest.scala
+++ b/core/src/test/scala/geotrellis/server/LayerExtentTest.scala
@@ -16,32 +16,24 @@
 
 package geotrellis.server
 
-import geotrellis.server.extent.SampleUtils
-
 import geotrellis.raster._
 import geotrellis.vector._
-import com.azavea.maml.error._
-import cats._
-import cats.effect._
-import cats.data.{NonEmptyList => NEL}
-import cats.implicits._
+import geotrellis.raster.resample.NearestNeighbor
 
 import org.scalatest._
 
-import scala.util.Random
 import scala.concurrent.ExecutionContext
-
 
 class LayerExtentTest extends FunSuite with Matchers {
   implicit val cs = cats.effect.IO.contextShift(ExecutionContext.global)
 
   test("ability to read a selected extent") {
-    val rt = ResourceTile("8x8.tif")
+    val rt = ResourceTile("8x8.tif", NearestNeighbor)
     val eval = LayerExtent.identity(rt)
     // We'll sample such that the bottom row (from 56 to 64) are excised from the result
     val sampled = eval(Extent(0, 1, 8, 8), CellSize(1, 1)).unsafeRunSync
     val sample = sampled.toOption.get.band(0).toArray()
-    val sampleSum = sample.fold(0)(_ + _)
+    val sampleSum = sample.sum
     assert(sampleSum == 1596, s"Expected sum of 1596, got $sampleSum")
   }
 }
--- a/ogc-example/src/main/resources/application.conf
+++ b/ogc-example/src/main/resources/application.conf
@@ -200,6 +200,7 @@ layers = {
         type = "simplesourceconf"
         name = "NLCD 2011",
         title = "National Land Cover Dataset 2011"
+        resample-method = "nearest-neighbor"
         sources = [
             "gt+s3://azavea-datahub/catalog?layer=nlcd-2011-epsg3857&zoom=13&band_count=1"
         ]
@@ -243,6 +244,7 @@ layers = {
         type = "simplesourceconf"
         name = "us-census-median-household-income"
         title = "US Sensus Median Household Income 20??"
+        resample-method = "nearest-neighbor"
 
         sources = [
             "gt+s3://azavea-datahub/catalog?layer=us-census-median-household-income-30m-epsg3857&zoom=12&band_count=1"
@@ -260,6 +262,7 @@ layers = {
         type = "simplesourceconf"
         name = "us-ned"
         title = "US NED"
+        resample-method = "nearest-neighbor"
         sources = [
             "gt+s3://azavea-datahub/catalog?layer=us-ned-tms-epsg3857&zoom=14&band_count=1"
         ]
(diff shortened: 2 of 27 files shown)


core/src/test/scala/geotrellis/server/LayerExtentTest.scala
```diff
@@ -16,32 +16,24 @@
...
-import scala.util.Random
 import scala.concurrent.ExecutionContext
 
-
 class LayerExtentTest extends FunSuite with Matchers {
   implicit val cs = cats.effect.IO.contextShift(ExecutionContext.global)
 
   test("ability to read a selected extent") {
-    val rt = ResourceTile("8x8.tif")
+    val rt = ResourceTile("8x8.tif", NearestNeighbor)
```
pomadchin: Hm, mb NN should be a default method everywhere? @CloudNiner

ogc-example/src/test/scala/geotrellis/server/HillshadeSpec.scala
```diff
@@ -0,0 +1,59 @@
...
+  implicit class TypeRefinement(self: Interpreted[Result]) {
+    def as[T: ClassTag]: Interpreted[T] = self match {
+      case Valid(r) => r.as[T]
+      case i @ Invalid(_) => i
+    }
+  }
+
+  // https://github.com/geotrellis/geotrellis-server/issues/150
+  describe("HillshadeSpec") {
+    ignore("RasterSource reproject hillshade") {
```
pomadchin: Ignore, since it is read from the S3 catalog. Mb we can think of a better test later (?)

ogc/src/main/scala/geotrellis/server/ogc/OgcSource.scala
```diff
@@ -51,6 +51,7 @@ trait OgcSource {
   def metadata: RasterMetadata
   def attributes: Map[String, String]
   def time: Option[ZonedDateTime]
+  def resampleMethod: ResampleMethod
```
CloudNiner: Is this part of the OGC definitions for "what an OGC layer is capable of" and does this attribute meet the requirements for inclusion defined in the docstring? Just double checking that this belongs as part of this trait so that we don't start to hang attributes off of OgcSource that make it less composable.

ogc-example/src/main/scala/geotrellis/server/ogc/conf/OgcSourceConf.scala
```diff
@@ -17,37 +17,37 @@
...
 }
 
 case class SimpleSourceConf(
   name: String,
   title: String,
   sources: List[String],
   defaultStyle: Option[String],
-  styles: List[StyleConf]
+  styles: List[StyleConf],
+  resampleMethod: ResampleMethod
```
CloudNiner: Should this default instead to "nearest-neighbor" so that existing confs are backwards compatible rather than forcing each user to make a choice about each layer they've defined? A default would also reduce the burden on new users when generating their first conf.

I don't feel strongly about it, just thinking through the implications of a change like this. If we leave it as is, we should definitely clarify in the changelog that this change is breaking for existing confs and point users to the list of available options and even perhaps what the implications of each option are, if we have that info available.

ogc-example/src/test/scala/geotrellis/server/HillshadeSpec.scala
CloudNiner: Ha. I was just about to comment that this test needs credentials. Beat me to it. 

In reality, doesn't this test belong in MAML or GeoTrellis since it doesn't have an GT server components at all? Perhaps one of those repos is better setup for these types of tests.

pomadchin: Probably should live eventually in GeoTrellis; wanted to capture it at least somewhere.

ogc/src/main/scala/geotrellis/server/ogc/OgcSource.scala
pomadchin: Yes, I think it is a 'what method for all resampling operation this raster source would use'; checkout also #229

ogc-example/src/main/scala/geotrellis/server/ogc/conf/OgcSourceConf.scala
pomadchin: I also didn't feel strong; see https://github.com/geotrellis/geotrellis-server/pull/233#discussion_r396420013 :D I'll add a default behavior than, since I also had a feeling about that but was not sure :+1:

CloudNiner: Oh I missed that comment -- :great-minds:. Adding a default sounds good, just leave the non-default new layer in the application.conf to demonstrate that it's possible. Maybe link to the non-default layer in the changelog so users can see an example at a glance?

ogc/src/main/scala/geotrellis/server/ogc/OgcSource.scala
CloudNiner: 👍

ogc-example/src/test/scala/geotrellis/server/HillshadeSpec.scala
CloudNiner: 👍 perhaps make an issue in locationtech/geotrellis? We might never get to it, but then its at least documented somewhere that its something we wanted to do...

ogc-example/src/main/scala/geotrellis/server/ogc/conf/OgcSourceConf.scala
pomadchin: @CloudNiner :+1:

ogc-example/src/main/resources/application.conf
```diff
@@ -334,6 +338,33 @@ layers = {
             }
         ]
     }
+    us-ned-hillshade = {
+      type = "mapalgebrasourceconf"
+      name = "us-ned-hillshade"
+      title = "US NED Hillshade"
+      resample-method = "bilinear"
```
pomadchin: `resample-method` is a new optional parameter, that's the usage example.

CloudNiner:
👍 still LGTM!

pomadchin:
Added an issue to move HillshadeSpecs from this repo into the GeoTrellis core. Mergin once CircleCI is happy!

PR #234, opened 2020-03-19
[WIP] Temporal improvements

### LINKED ISSUE #219: GeoTrellisRasterSource temporal build improvements

CloudNiner:
## Overview

**My comments are in the git log for the HEAD commit on this branch. Additional context there.**

This PR addresses the major concerns raised in #219:
- [x] Can we revert to upstream GeoTrellisRasterSource? Yes -- we'll need to add the time related changes to upstream though
- [x] Can we minimize the number of RasterSource instantiations we do for GeoTrellis layers? Yes -- Now 2. One when we build the conf on server start, one every time we go to read from the layer in a GetMap or equivalent call
- [x] Can we defer RasterSource instantiation until just before we're ready to read data? Yes, see bullet above

### Checklist

## Testing Instructions

It should still be possible to load and view both application.conf. It should still be possible to load application-spacetimekey.conf and view the GetCapabilities or equivalent views for WMS/WCS manually. GetMap or equivalent views will still time out due to excessive reads of the underlying test rastersource.

Closes #219

### DIFF
--- a/ogc-example/src/main/resources/application-spacetimekey.conf
+++ b/ogc-example/src/main/resources/application-spacetimekey.conf
@@ -187,16 +187,10 @@ wmts = {
 
 layers = {
   spacetimetest = {
-    type = "simplesourceconf"
+    type = "rastersourceconf"
     name = "spacetimetest"
     title = "Landsat 8 OGC Temporal SpaceTimeKey Test"
-    source = {
-      type = "geotrellis"
-      catalog-uri = "s3://geotrellis-test-non-public/spacetimekey-test"
-      layer = "spacetimekey-test"
-      zoom = 14
-      band-count = 1
-    }
+    source = "gt+s3://geotrellis-test-non-public/spacetimekey-test?layer=spacetimekey-test&zoom=14&band_count=1"
     styles = [
       {
         name = "red-to-blue"
--- a/ogc/src/main/scala/geotrellis/server/ogc/OgcSource.scala
+++ b/ogc/src/main/scala/geotrellis/server/ogc/OgcSource.scala
@@ -16,24 +16,25 @@
 
 package geotrellis.server.ogc
 
-import geotrellis.server.extent.SampleUtils
-import geotrellis.server.ogc.wms._
-import geotrellis.server.ogc.style._
-
-import geotrellis.raster._
-import geotrellis.vector.{Extent, ProjectedExtent}
-import geotrellis.proj4.CRS
-
-import com.azavea.maml.ast._
-import cats.data.{NonEmptyList => NEL}
-import opengis.wms.BoundingBox
-
 import java.time.ZonedDateTime
+
+import cats.data.{NonEmptyList => NEL}
+import cats.syntax.option._
+import com.azavea.maml.ast.Expression
+import geotrellis.proj4.CRS
+import geotrellis.raster._
+import geotrellis.server.extent.SampleUtils
+import geotrellis.server.ogc.style._
+import geotrellis.server.ogc.wms.CapabilitiesView
+import geotrellis.store.{GeoTrellisPath, GeoTrellisRasterSourceLegacy, LayerHeader}
+import geotrellis.vector.{Extent, ProjectedExtent}
+import jp.ne.opt.chronoscala.Imports._
+import opengis.wms.BoundingBox
 
 /**
  * This trait and its implementing types should be jointly sufficient, along with a WMS 'GetMap'
  *  (or a WMTS 'GetTile' or a WCS 'GetCoverage' etc etc) request to produce a visual layer
- *  (represented more fully by the [[OgcLayer]] hierarchy.
+ *  (represented more fully by [[OgcLayer]].
  *  This type represents *merely* that there is some backing by which valid OGC layers
  *  can be realized. Its purpose is to provide the appropriate level of abstraction for OGC
  *  services to conveniently reuse the same data about underlying imagery
@@ -50,10 +51,30 @@ trait OgcSource {
   def nativeCrs: Set[CRS]
   def metadata: RasterMetadata
   def attributes: Map[String, String]
-  def time: Option[ZonedDateTime]
   def resampleMethod: ResampleMethod
+  def timeInterval: Option[OgcTimeInterval]
 
   def nativeProjectedExtent: ProjectedExtent = ProjectedExtent(nativeExtent, nativeCrs.head)
+}
+
+trait RasterOgcSource extends OgcSource {
+  def source: RasterSource
+
+  def extentIn(crs: CRS): Extent = {
+    val reprojected = source.reproject(crs)
+    reprojected.extent
+  }
+
+  def bboxIn(crs: CRS): BoundingBox = {
+    val reprojected = source.reproject(crs)
+    CapabilitiesView.boundingBox(crs, reprojected.extent, reprojected.cellSize)
+  }
+
+  lazy val nativeRE: GridExtent[Long]      = source.gridExtent
+  lazy val nativeCrs: Set[CRS]             = Set(source.crs)
+  lazy val nativeExtent: Extent            = source.extent
+  lazy val metadata: RasterMetadata        = source.metadata
+  lazy val attributes: Map[String, String] = metadata.attributes
 }
 
 /**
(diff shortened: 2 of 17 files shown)


ogc/src/main/scala/geotrellis/server/ogc/source/GeoTrellisOgcSource.scala
```diff
@@ -0,0 +1,64 @@
...
+    timeMetadataKey: String = "times"
+) extends RasterOgcSource {
+
+  private val dataPath = GeoTrellisPath.parse(sourceUri)
+
+  // If layer is temporally enabled, default time is the first in the layer
+  lazy val source =
+    GeoTrellisRasterSourceLegacy(dataPath, timeInterval.map(_.start))
+
+  lazy val times: Option[NEL[ZonedDateTime]] = {
```
echeipesh: Optional NonEmptyList, so is that like a list?
Didn't trace the semantics of this this but the signature made me laugh.

pomadchin: `List` 💯

ogc/src/main/scala/geotrellis/server/ogc/wms/WmsModel.scala
```diff
@@ -16,62 +16,78 @@
...
+            p.styles.headOption
+              .filterNot(_.isEmpty)
+              .orElse(source.defaultStyle)
+          val style: Option[OgcStyle] =
+            styleName.flatMap { name =>
+              source.styles.find(_.name == name)
+            }
+          source match {
+            case MapAlgebraSource(name,
+                                    title,
```
pomadchin: How do you feel about such formating? Looks a bit unreadble to me ._.

Also it looks like the old `for` syntax was okay; if you don't like it how about smth like:
image

ogc/src/main/scala/geotrellis/server/ogc/source/GeoTrellisOgcSource.scala
```diff
@@ -0,0 +1,64 @@
...
+            case OgcTimeInterval(start, Some(end), _) => start <= t && t < end
+            case _                                    => false
+          }
+        }
+        .fold(source)(sourceForTime(_))
+    }
+  }
+
+  def sourceForTime(time: ZonedDateTime): GeoTrellisRasterSourceLegacy =
+    GeoTrellisRasterSourceLegacy(dataPath, Some(time))
```
pomadchin: It looks like there still a new raster sources for a time point is created. Have you decided to move it into a separate PR?

ogc/src/main/scala/geotrellis/server/ogc/source/GeoTrellisOgcSource.scala
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  def sourceForTime(timeInterval: OgcTimeInterval): GeoTrellisRasterSourceLegacy = {
+    // I'd love some suggestions for how to write this better...
+    times.fold(source) { timeList =>
+      timeList
+        .find { t =>
+          timeInterval match {
+            case OgcTimeInterval(start, None, _)      => start == t
+            case OgcTimeInterval(start, Some(end), _) => start <= t && t < end
```
pomadchin: Find will return on the first occurrence, is it what we expect?

ogc/src/main/scala/geotrellis/server/ogc/source/GeoTrellisOgcSource.scala
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  lazy val timeInterval: Option[OgcTimeInterval] = times.map {
+    case NEL(head, Nil)  => OgcTimeInterval(head)
+    case NEL(head, tail) => OgcTimeInterval(head, Some(tail.max), None)
+  }
+
+  def sourceForTime(timeInterval: OgcTimeInterval): GeoTrellisRasterSourceLegacy = {
+    // I'd love some suggestions for how to write this better...
```
pomadchin: It would be smth like 

```
def sourceForTime(timeInterval: OgcTimeInterval): GeoTrellisRasterSourceLegacy = 
    if(times.isEmpty) source
    else times.find { t =>
      timeInterval match {
        case OgcTimeInterval(start, None, _) => start == t
        case OgcTimeInterval(start, Some(end), _) => start <= t && t < end
        case _ => false
      }
    }.fold(source)(sourceForTime)
```

After you'll change the `times` signature.

CloudNiner:
>  Is it expected that in this PR there was nothing done to optimize GeoTrellisRasterSources creation

So the change to address that is here: https://github.com/geotrellis/geotrellis-server/pull/234/files#diff-fb4892c4b98bbccefb508b128e997082R38

Previously when the server started we were creating a new RasterSource for all possible times in the GtRasterSource (via the now removed `build()` function) + the parent raster source. Now we create the parent RasterSource on server start to get the valid times out of it, and then only create a single RasterSource valid for the request time when a user makes a temporal request. Requests with no temporal component just use the parent RasterSource defined at `GeoTrellisOgcSource.source`. This seems like the best we can do...at some point we have to create a raster source valid at the appropriate time in order to request the data. Is there some further optimization we can make?

ogc/src/main/scala/geotrellis/server/ogc/source/GeoTrellisOgcSource.scala
CloudNiner: LGTM, updated

ogc/src/main/scala/geotrellis/server/ogc/source/GeoTrellisOgcSource.scala
CloudNiner: Switched to `List`

ogc/src/main/scala/geotrellis/server/ogc/source/GeoTrellisOgcSource.scala
CloudNiner: Yeah. Since we can only render a raster valid for a single time, we just pick the first one in the list that matches the interval. I added a `.sorted` to `lazy val times` so that we always pick the time closest to the start interval if there are multiple matches. That way there's some consistency.

ogc/src/main/scala/geotrellis/server/ogc/source/GeoTrellisOgcSource.scala
CloudNiner: I _think_ this is the same comment [as below](https://github.com/geotrellis/geotrellis-server/pull/234#issuecomment-603477419). Correct me if I'm wrong?

ogc/src/main/scala/geotrellis/server/ogc/wms/WmsModel.scala
CloudNiner: Oof. Yeah the formatting here sucks. I hate that the intellij scalafmt formatter defaults to indenting parameters like that.

Updated

ogc/src/main/scala/geotrellis/server/ogc/OgcSource.scala
```diff
@@ -75,14 +69,73 @@ case class SimpleSource(
...
+  lazy val source =
+    GeoTrellisRasterSourceLegacy(dataPath, timeInterval.map(_.start))
+
+  lazy val isTemporal: Boolean = !times.isEmpty
+
+  lazy val times: List[ZonedDateTime] = {
+    val layerId = source.dataPath.layerId
+    val header = source.attributeStore.readHeader[LayerHeader](layerId)
+    if (header.keyClass.contains("SpaceTimeKey")) {
+      source.attributeStore.read[List[ZonedDateTime]](layerId, timeMetadataKey).sorted
```
pomadchin: Add an import `geotrellis.store.query.CirceCompat._` in all places you use temporal circe codecs. The sotry behind that is that Scala 2.12 circe handles temporal codecs in a differnet way (they are supproted by default) and in early circe versions such codecs lived in a separate circe module.

pomadchin: @CloudNiner ^ this is a comment that explains why Scala 2.11 builds are failing.

ogc/src/main/scala/geotrellis/server/ogc/OgcSource.scala
```diff
@@ -77,14 +70,75 @@ case class SimpleSource(
...
+  timeMetadataKey: String = "times"
+) extends RasterOgcSource {
+
+  private val dataPath = GeoTrellisPath.parse(sourceUri)
+
+  // If layer is temporally enabled, default time is the first in the layer
+  lazy val source =
+    GeoTrellisRasterSourceLegacy(dataPath, timeInterval.map(_.start))
+
+  lazy val isTemporal: Boolean = !times.isEmpty
```
pomadchin: Actually there is a `.nonEmpty` function :D

ogc-example/src/main/resources/application-spacetimekey.conf
```diff
@@ -190,13 +190,7 @@ layers = {
...
     name = "spacetimetest"
     title = "Landsat 8 OGC Temporal SpaceTimeKey Test"
-    source = {
-      type = "geotrellis"
-      catalog-uri = "s3://geotrellis-test-non-public/spacetimekey-test"
-      layer = "spacetimekey-test"
-      zoom = 14
-      band-count = 1
-    }
+    source = "gt+s3://geotrellis-test-non-public/spacetimekey-test?layer=spacetimekey-test&zoom=14&band_count=1"
```
pomadchin: Also it looks like that `ogc-example/src/main/resources/application.conf` has to be changed and tested before the merge.

CloudNiner: Yup. Addressed.

ogc/src/main/scala/geotrellis/server/ogc/OgcSource.scala
pomadchin: Ok, [here is the problem](https://github.com/locationtech/geotrellis/blob/0e60b860825d289f7283bea9683f3ca4b2f3c4c5/store/src/main/scala/geotrellis/store/json/Implicits.scala#L95-L100) we need to remove it from the GeoTrellis core.

pomadchin:
Create a new GeoTrellis issue https://github.com/locationtech/geotrellis/issues/3208

ogc/src/main/scala/geotrellis/server/ogc/OgcSource.scala
CloudNiner: Issue here resolved by removing the CirceCompat import from GeoTrellisRasterSourceLegacy which pulls a ZonedDateTime decoder from the package listed in the link above.

CloudNiner:
Whoops, missed the `rastersourceconf` change. I updated CHANGELOG to note that before I merged.

pomadchin:
Docs live here https://github.com/geotrellis/geotrellis-server/blob/develop/ogc-example/docs/conf.md#temporal-layers (adding this link for a better discoverability)

Issue #239, opened 2020-03-25
Update locationtech repo urls


CloudNiner:
To use repo.eclipse.org, e.g. https://repo.eclipse.org/content/groups/snapshots/org/locationtech/geotrellis/geotrellis-s3_2.12/

PR #240, opened 2020-03-26
Use Eclipse repositories instead of locationtech

pomadchin:
Locationtech repos would go away soon, this PR switches us to eclipse repos.

Closes https://github.com/geotrellis/geotrellis-server/issues/239


CloudNiner:
👍

PR #241, opened 2020-03-27
Add temporal layers docs

pomadchin:
This PR adds some information about how to add temporal attributes into the already ingested layer.


CloudNiner:
👍 nice addition!

PR #245, opened 2020-03-30
Add ColorRampStyle.clampWithColor

### LINKED ISSUE #220: A 'Clamped' option for color-ramp definition with min-render/max-render boundaries
`min-render` and `max-render` properties allow service administrators to define rendering boundaries for color-ramp definition. By default, values outside those boundaries are rendered as transparent pixels. For instance with the following configuration, any NDVI values above 0.5 or below -0.5 will be rendered as transparent. 
```
type = "colorrampconf"
colors = ${color-ramps.ndvi}
min-render = -0.5
max-render = 0.5
```
It it might be interesting to offer to clamp values outside those ranges to color-ramp lowest/highest colors.

CloudNiner:
## Overview

Adds option `clamp-with-color` to `colorrampstyle` that renders values outside the requested min/max render range with the min/max color values from the ramp rather than transparent pixels. 

Default is false to maintain backwards
compatibility for existing configurations

I'm taking naming suggestions for a better name for the new config option. A succinct descriptive name remains elusive.

### Checklist

### Demo

With the following change applied:
```diff
diff --git a/ogc-example/src/main/resources/application.conf b/ogc-example/src/main/resources/application.conf
index 50209b1..a2170d1 100644
--- a/ogc-example/src/main/resources/application.conf
+++ b/ogc-example/src/main/resources/application.conf
@@ -260,13 +260,17 @@ layers = {
         name = "us-ned"
         title = "US NED"
         source = "gt+s3://azavea-datahub/catalog?layer=us-ned-tms-epsg3857&zoom=14&band_count=1"
+        default-style = "Elevation Ramp"
         styles = [
           {
             type = "colorrampconf"
             name = "Elevation Ramp"
             title = "Elevation Ramp"
-            colors = ${color-ramps.elevation}
+            colors = ${color-ramps.red-to-blue}
             stops = 90
+            min-render = 1000
+            max-render = 3000
+            clamp-with-color = true
           }
         ]
     }
```

This layer will render like this for `clamp-with-color = true`:
Screen Shot 2020-03-30 at 2 50 28 PM

And like this for `clamp-with-color = false` (the default, should match the rendering on `develop`):
Screen Shot 2020-03-30 at 2 51 12 PM

## Testing Instructions

Apply the diff shown in Demo to your application.conf and load the layer. Ensure that it looks the same on develop and with `clamp-with-color = false` and that colors are rendered instead of transparent pixels with the new option enabled.

Closes #220

### DIFF
--- a/ogc-example/src/main/scala/geotrellis/server/ogc/conf/StyleConf.scala
+++ b/ogc-example/src/main/scala/geotrellis/server/ogc/conf/StyleConf.scala
@@ -36,10 +36,11 @@ final case class ColorRampConf(
   stops: Option[Int],
   minRender: Option[Double],
   maxRender: Option[Double],
+  clampWithColor: Boolean = false,
   legends: List[LegendModel] = Nil
 ) extends StyleConf {
   def toStyle: OgcStyle =
-    ColorRampStyle(name, title, colors, stops, minRender, maxRender, legends)
+    ColorRampStyle(name, title, colors, stops, minRender, maxRender, clampWithColor, legends)
 }
 
 /** Styling in which both a color scheme and the data-to-color mapping is known */
--- a/ogc/src/main/scala/geotrellis/server/ogc/style/ColorRampStyle.scala
+++ b/ogc/src/main/scala/geotrellis/server/ogc/style/ColorRampStyle.scala
@@ -24,7 +24,6 @@ import geotrellis.raster.render.{ColorMap, ColorRamp}
 import geotrellis.raster.render.jpg.JpgEncoder
 import geotrellis.util.np.linspace
 
-
 case class ColorRampStyle(
   name: String,
   title: String,
@@ -32,6 +31,7 @@ case class ColorRampStyle(
   stops: Option[Int],
   minRender: Option[Double],
   maxRender: Option[Double],
+  clampWithColor: Boolean = false,
   legends: List[LegendModel] = Nil
 ) extends OgcStyle {
   lazy val logger = org.log4s.getLogger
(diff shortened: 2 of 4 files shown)


ogc-example/src/main/scala/geotrellis/server/ogc/conf/StyleConf.scala
```diff
@@ -36,10 +36,10 @@ final case class ColorRampConf(
   stops: Option[Int],
   minRender: Option[Double],
   maxRender: Option[Double],
+  clampWithColor: Boolean = false,
```
CloudNiner: An alternative to this single option would be two independent additional options:
- `maxRenderColor: Int = 0x00000000`
- `minRenderColor: Int = 0x00000000`

Which would serve the same purpose, but require that the user inspect the color ramp for the default case where the user wants the min/max colors to match those of the ramp.

In any case, the implementation here is explicitly what's requested in the issue.

ogc/src/main/scala/geotrellis/server/ogc/style/ColorRampStyle.scala
```diff
@@ -59,21 +59,22 @@ case class ColorRampStyle(
     // The number of stops between each provided break
     val numStops: Int = stops.getOrElse(colorRamp.colors.length)
 
-    // The colors, interpolated (invisible added for values below our minimum break)
-    val interpolatedColors: Vector[Int] =
-      0x00000000 +: colorRamp.stops(numStops)
+    // The colors, interpolated (colors added at start and end for out of bounds values)
+    val minColor = if (clampWithColor) colorRamp.colors.head else 0x00000000
+    val maxColor = if (clampWithColor) colorRamp.colors.last else 0x00000000
```
pomadchin: Why should the maxColor be set to `0x00000000`?

ogc/src/main/scala/geotrellis/server/ogc/style/ColorRampStyle.scala
```diff
@@ -59,21 +59,22 @@ case class ColorRampStyle(
...
+    val interpolatedColors: Vector[Int] = minColor +: colorRamp.stops(numStops) :+ maxColor
 
     val interpolatedBreaks: Array[Double] =
-      breaks(hists, interpolatedColors.length)
+      breaks(hists, interpolatedColors.length) :+ Double.MaxValue
```
pomadchin: Hm, why does the `Double.MaxValue` in the interpolation breaks?

pomadchin:
Also, let's add one example into the `application.conf` file?

ogc/src/main/scala/geotrellis/server/ogc/style/ColorRampStyle.scala
echeipesh: The way I'm reading it is if the rendering is not clamped to max color value, that is forced to use it, it should be transparent.

ogc/src/main/scala/geotrellis/server/ogc/style/ColorRampStyle.scala
CloudNiner: I _think_ if I'm remembering correctly that without this colors above the max-render value would always render transparent.

CloudNiner: Verified. If you want to render a color for each bin, you need a bin that represents everything greater than the max-render value. If I remove this extra appended value then values between max-render and Double.MaxValue aren't represented by a bin and are therefore always rendered as transparent pixels.

moradology: I mention this only for the sake of airing the knowledge and think your solution is probably nicer.

An alternate way to control rendering outside of the binds provided within a `ColorMap` is to initialize the `ColorMap` with a custom `ColorMap.Options` object:
```
val breaks = ???
val ramp = ???
val fallbackRGB = ???
val options = ColorMap.Options.DEFAULT.copy(fallbackColor = fallbackRGB)
ramp.toColorMap(breaks, ramp, ColorMap
```

moradology:
Looks good to me; +1

PR #251, opened 2020-04-10
RGB Layers support

### LINKED ISSUE #249: RGB styling configuration
After implementing [MAML RGB Composite operation](https://github.com/geotrellis/maml/issues/114) and all the nessesary color correction nodes (at least rescaling) we'd have to expose this functionality into the GeoTrellis server. 

The parameters used for the default scaling should be possible to set through the application.conf file (potentially through the MAML expression). It should be also possible to override set parameters through the extended GetMap / GetCoverage get arguments.

pomadchin:
## Overview

This PR bumps MAML version up and adds test RGB layers and Extended user parameters

### Checklist

### Demo

image

Closes #249


pomadchin:
@jfbourgon  I think I'll merge it it for now; if there would be extra follow up comments, we'll create new issues :)

PR #267, opened 2020-06-04
Clarify the FilterSearch Semigroup behavior

pomadchin:
## Overview

This PR cleanps SearchFilters Semigroup

### DIFF
--- a/stac-example/src/main/scala/geotrellis/stac/api/SearchFilters.scala
+++ b/stac-example/src/main/scala/geotrellis/stac/api/SearchFilters.scala
@@ -25,10 +25,12 @@ import io.circe.generic.semiauto._
 import io.circe.syntax._
 import higherkindness.droste.{Algebra, scheme}
 import geotrellis.vector._
-import cats.syntax.option._
 import cats.Semigroup
-import cats.instances.option._
+import cats.syntax.option._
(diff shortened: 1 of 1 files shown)


stac-example/src/main/scala/geotrellis/stac/api/SearchFilters.scala
```diff
@@ -55,14 +57,28 @@ object SearchFilters {
   implicit val geometryIntersectionSemigroup: Semigroup[Geometry] =
     Semigroup.instance { (left, right) => left.intersection(right) }
 
+  implicit val bboxIntersectionSemigroup: Semigroup[Bbox] =
+    Semigroup.instance { (left, right) =>
```
CloudNiner, line 61: Is this `Semigroup.instance` a shortcut to avoid using the anonymous class syntax?

pomadchin: Exactly!

CloudNiner: Cool, gonna use that over in the other semigroup PR if I can. 

The more of this functional stuff I do, the more I think Scala needs more ways to do the same thing.... _/s_

PR #269, opened 2020-06-04
Implement Semigroup for RepositoryM

CloudNiner:
## Overview

Simplifies combination of Repositories and avoids having to construct wrapper repos for lists of repos.

Also defaults RepositoryM.store definition in terms of RepositoryM.find


core/src/main/scala/geotrellis/store/query/Repository.scala
```diff
@@ -16,11 +16,24 @@
...
 }
 
-trait Repository[G[_], T] extends RepositoryM[Id, G, T]
+object RepositoryM {
+  implicit def semigroupRepositoryM[M[_]: FlatMap: Functor, G[_]: SemigroupK, T]
+    : Semigroup[RepositoryM[M, G, T]] =
+    (x: RepositoryM[M, G, T], y: RepositoryM[M, G, T]) =>
+      new RepositoryM[M, G, T] {
+
+        override def find(query: Query): M[G[T]] =
```
pomadchin, line 36: what do you think about `convention` I'm trying to follow? 
to use `override` only in places where it overrides the default behavior? (like in these `override val store` cases).

core/src/main/scala/geotrellis/store/query/Repository.scala
```diff
@@ -16,11 +16,24 @@
...
 package geotrellis.store.query
 
-import cats.Id
+import cats.{FlatMap, Functor, Semigroup, SemigroupK}
+import cats.syntax.functor._
+import cats.syntax.flatMap._
+import cats.syntax.semigroupk._
+import geotrellis.store.query
 
 trait RepositoryM[M[_], G[_], T] {
```
pomadchin, line 25: The file name should be `RepositoryM` now.

core/src/main/scala/geotrellis/store/query/Repository.scala
```diff
@@ -16,11 +16,24 @@
...
 
-trait Repository[G[_], T] extends RepositoryM[Id, G, T]
+object RepositoryM {
+  implicit def semigroupRepositoryM[M[_]: FlatMap: Functor, G[_]: SemigroupK, T]
+    : Semigroup[RepositoryM[M, G, T]] =
+    (x: RepositoryM[M, G, T], y: RepositoryM[M, G, T]) =>
+      new RepositoryM[M, G, T] {
+
+        override def find(query: Query): M[G[T]] =
+          x.find(query).flatMap(g => y.find(query).map(_ <+> g))
```
pomadchin, line 37: what do you think about the ordering of this combine? mb it should be `g <+> _`?

core/src/main/scala/geotrellis/store/query/Repository.scala
```diff
@@ -16,11 +16,24 @@
...
-  def store: M[G[T]]
+  def store: M[G[T]] = find(query.all)
   def find(query: Query): M[G[T]]
 }
 
-trait Repository[G[_], T] extends RepositoryM[Id, G, T]
+object RepositoryM {
+  implicit def semigroupRepositoryM[M[_]: FlatMap: Functor, G[_]: SemigroupK, T]
+    : Semigroup[RepositoryM[M, G, T]] =
+    (x: RepositoryM[M, G, T], y: RepositoryM[M, G, T]) =>
```
pomadchin, line 33: Yea, this syntax is not supported by Scala 2.11, as a workaround, you can use the following syntax:

```scala
Semigroup.instance { (left, right) =>
  new RepositoryM[M, G, T] { ... }
}
```

core/src/main/scala/geotrellis/store/query/Repository.scala
```diff
@@ -16,11 +16,24 @@
...
 
 trait RepositoryM[M[_], G[_], T] {
-  def store: M[G[T]]
+  def store: M[G[T]] = find(query.all)
   def find(query: Query): M[G[T]]
 }
 
-trait Repository[G[_], T] extends RepositoryM[Id, G, T]
+object RepositoryM {
+  implicit def semigroupRepositoryM[M[_]: FlatMap: Functor, G[_]: SemigroupK, T]
```
pomadchin, line 31: `: FlatMap: Functor` is a `Monad` definition here

core/src/main/scala/geotrellis/store/query/Repository.scala
pomadchin: AH, how do you feel about doing it through [Applicative](https://typelevel.org/cats/typeclasses/applicative.html#applicatives-compose)? What we really need is independent composition, but `Monad` gives us capablities to chain operations.

core/src/main/scala/geotrellis/store/query/Repository.scala
pomadchin: Oooooh, probably we don't need `Functor`. `FlatMap` provides you a [map2](https://typelevel.org/cats/api/cats/FlatMap.html#map2[A,B,Z](fa:F[A],fb:F[B])(f:(A,B)=%3EZ):F[Z]) function, so we can use that in composition.

pomadchin: ^ I didn't try the above so i can be wrong

core/src/main/scala/geotrellis/store/query/Repository.scala
CloudNiner: Works for me. Updated. Looks like this was the only spot I did this that I could find.

core/src/main/scala/geotrellis/store/query/Repository.scala
CloudNiner: That does it...and is so much cleaner. That's a nifty shortcut.

PR #271, opened 2020-06-05
Receive GPG key while publishing artifacts

ColeKettler:
## Overview

This retrieves an up-to-date copy of the Geotrellis public GPG key during the `cipublish` build stage.

We've pushed a new signature to extend the expiration date on the public key that we've uploaded to public keyservers. The CircleCI build will now retrieve the latest copy of the public key, allowing us to renew the key in the future using the same procedure.

### Checklist

## Testing Instructions

- Verify that updated and valid Geotrellis GPG key is retrieved and verified during `cipublish`.
  - Verified by @rbreslow during pairing session.

Connects azavea/operations#446

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,8 @@ aliases:
     - run:
         name: "Import signing key"
         command: |
+          gpg --keyserver keyserver.ubuntu.com \
+            --recv-keys 0x13E9AA1D8153E95E && \
           echo "${GPG_KEY}" | base64 -d > signing_key.asc && \
           gpg --batch \
             --import signing_key.asc
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,7 +85,6 @@ workflows:
                 - develop
                 - /release\/.*/
                 - /hotfix\/.*/
-                - feature/cek/receive-gpg-key
 
 jobs:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -25,10 +25,10 @@ aliases:
     - run:
         name: "Import signing key"
         command: |
+          gpg --keyserver keyserver.ubuntu.com \
+            --recv-keys 0x13E9AA1D8153E95E && \
           echo "${GPG_KEY}" | base64 -d > signing_key.asc && \
-          gpg --batch \
-            --passphrase "${GPG_PASSPHRASE}" \
```
ColeKettler, line 30: Because we're not using passphrase for this key, the `--batch` and `--passphrase` options are unnecessary.

.circleci/config.yml
```diff
@@ -25,10 +25,10 @@ aliases:
     - run:
         name: "Import signing key"
         command: |
+          gpg --keyserver keyserver.ubuntu.com \
```
rbreslow, line 28: It's worth surfacing that we're specifying the Ubuntu keyserver here to mirror the [approach taken](https://github.com/jodersky/sbt-gpg/blob/master/.ci/build) by @jodersky. Through testing, I've found that this keyserver was updated more quickly than the default.

Issue #259, opened 2020-05-20
Support for Time lists in WMS GetCapabilities


jfbourgon:
GeoTrellis-Server is currently providing information about the available Time dimension of a temporal layer using only a Time interval syntax but the OGC WMS 1.3.0 specifications (Annex D) also allows to display time dimension using a Time list syntax.

For instance, the following Time dimension could be displayed as :
```
<Dimension default="2014-07-01T12:00Z" units="ISO8601" name="time">2014-07-01T12:00:00Z,2015-07-01T12:00:00Z,2016-07-01T12:00:00Z</Dimension>
```
instead of only
```
<Dimension default="2014-07-01T12:00Z" units="ISO8601" name="time">2014-07-01T12:00:00Z/2016-07-01T12:00:00Z</Dimension>
```

Ideally, the service administrator could have a way to select the preferable syntax.

PR #279, opened 2020-06-16
Expose STAC temporal metadata

### LINKED ISSUE #260: Default attribute should not be used to serve response inside Time extent
According to the WMS 1.3.0 specifications, the default attribute defined on the Timedimension should only be used when the TIME parameter is omitted.

> **Default field**: Attribute indicating default value that will be used if GetMap request does not specify a value. If attribute is absent, then shall respond with a service exception if request
does not include a value for that dimension.

Current implementation serves the default Time for any non existent Time period requested between the Time Extent boundaries defined in the WMS GetCapabilities. 

For instance, based on the following Time Extent:
```scala
<Layer queryable="false">
<Name>red</Name>
<Title>Red band of LandSat Mosaic</Title>
...
<Dimension default="2014-07-01T12:00Z" units="ISO8601" name="time">2014-07-01T12:00:00Z/2016-07-01T12:00:00Z</Dimension>
...
</Layer>
```
with following Time period available in the source dataset:
```scala
<TemporalDomain>
<gml:TimePosition>2014-07-01T12:00:00Z</gml:TimePosition>
<gml:TimePosition>2015-07-01T12:00:00Z</gml:TimePosition>
<gml:TimePosition>2016-07-01T12:00:00Z</gml:TimePosition>
</TemporalDomain>
```

GetMap requests will throw BadRequest ServiceException for TIME=2017-07-01T12:00:00Z (value outside of the provided Time extent) but will serve the Default for TIME=2015-09-01T12:00:00Z (non existent value located inside the provided Time extent)

The expected behaviour would be to raise a ServicException in both cases.

Another option would be to implement support for the nearest attribute that would inform clients that the nearest value would be served if the request is not exactly matching one of the existing time period.

pomadchin:
## Overview

This PR adds exposes stac temporal extent thought the temporal metadata (and covers #259 at the same time):

image

It also fixes #260:

* (works) http://localhost:9000/?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&BBOX=43.81618476019971808%2C-79.40248489379884234%2C43.94325769600655462%2C-79.15603855282134305&CRS=EPSG%3A4326&WIDTH=1103&HEIGHT=570&LAYERS=markham&STYLES=&FORMAT=image/png&DPI=72&MAP_RESOLUTION=72&FORMAT_OPTIONS=dpi%3A72&TRANSPARENT=TRUE&time=1960-01-01T12:00:00Z 
* (returns an empty tile) http://localhost:9000/?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&BBOX=43.81618476019971808%2C-79.40248489379884234%2C43.94325769600655462%2C-79.15603855282134305&CRS=EPSG%3A4326&WIDTH=1103&HEIGHT=570&LAYERS=markham&STYLES=&FORMAT=image/png&DPI=72&MAP_RESOLUTION=72&FORMAT_OPTIONS=dpi%3A72&TRANSPARENT=TRUE&time=1961-01-01T12:00:00Z

### Checklist

Closes https://github.com/geotrellis/geotrellis-server/issues/259
Closes https://github.com/geotrellis/geotrellis-server/issues/260


ogc/src/main/scala/geotrellis/server/ogc/OgcSource.scala
```diff
@@ -90,9 +90,30 @@ case class SimpleSource(
   defaultStyle: Option[String],
   styles: List[OgcStyle],
   resampleMethod: ResampleMethod,
-  overviewStrategy: OverviewStrategy
+  overviewStrategy: OverviewStrategy,
+  timeMetadataKey: Option[String]
 ) extends RasterOgcSource {
-  val timeInterval: Option[OgcTimeInterval] = None
+  lazy val timeInterval: Option[OgcTime] =
```
pomadchin, line 96: Instead of options I’ll add the OgcTimeEmpty node

ogc/src/main/scala/geotrellis/server/ogc/wms/WmsParams.scala
```diff
@@ -108,8 +115,7 @@ object WmsParams {
           val height =
             params.validatedParam[Int]("height", { s => Try(s.toInt).toOption })
 
-          val time = params.validatedOptionalParam("time")
-            .map(option => option.map(OgcTimeInterval.fromString))
+          val time = params.validatedTemporalPosition("time")
```
CloudNiner, line 118: It looks like we've lost the ability to support WMS 1.3 queries that use the TimeInterval format since this method only parses results into OgcTimePositions or OgcTimeEmpty.

Querying by list of times and a single time interval look like they should both be supported, no? 

> OGC WMS 1.3.0 specifications (Annex D) **also** allows to display time dimension using a Time list syntax

ogc/src/main/scala/geotrellis/server/ogc/OgcTime.scala
```diff
@@ -17,9 +17,53 @@
...
+    case (l: OgcTimeInterval, r: OgcTimeInterval) => l |+| r
+    case (l: OgcTimePositions, _: OgcTimeEmpty.type) => l
+    case (l: OgcTimeInterval, _: OgcTimeEmpty.type) => l
+    case (_: OgcTimeEmpty.type, r: OgcTimePositions) => r
+    case (_: OgcTimeEmpty.type, r: OgcTimeInterval) => r
+    case (l, _) => l
+  }
+}
+
+final case class OgcTimePositions(list: NonEmptyList[ZonedDateTime]) extends OgcTime {
```
CloudNiner, line 48: The comma separated format described in #259 could be described by `List[OgcTimeInterval]` where only the `start` property is set for each `OgcTimeInterval`. In effect, `List[OgcTimePositions]` is the same as `List[OgcTimeInterval(start, None, None)]`, and we should be able to parse a string containing comma separated single times, comma separated intervals and a single interval all into the same `List[OgcTimeInterval]` type and then render them back out to XML based on the specifics of the WCS or WMS spec on the other side.

As a result, I'm curious as to why we now have two diverging types `OgcTimePositions` and `OgcTimeInterval`? What am I missing where the existing types weren't sufficient? I'm trying to avoid us adding additional code complexity where possible for an already complex topic.

pomadchin: @CloudNiner I decided to add  richer semantics to simplify the code.

pomadchin: `OgcTime` is an ADT now, wich can be an interval or can be a list of times that probably are not an interval. For sure Positions can be represented as `OgcTimeInterval(start, None, None)` but I don't think it is much more readable and the type can add semantics.

CloudNiner: Gotcha. Then in the absence of this change giving us some feature that we didn't have before, I have to disagree with this change. 

IMO it has cost us maintainability in that as a programmer looking to work on this library I have to unpack when and where I should use each OgcTime type when. And if I need a new type I have to add it everywhere and have a strong understanding what the code is doing at each point to implement correctly. We also are missing the case where a list of OgcTime could contain position and interval intermixed -- the implementation of that would be an additional OgcTime as `List[OgcTimeInterval]` which is just a general case of all the currently implemented OgcTime types.

A better way to apply some sort of ADT here would be to do it at the individual service level. For example, if WCS wants to explicitly support a lists of positions, then have something like a WCSTimePositions specific to that service rather than complicating the current `List[OgcTimeInterval]` data structure that is defined as described in the ogc spec.

Or, if the core change here is that we need an `OgcSource` to support a list of times, then we should just change `def timeInterval: Option[OgcTimeInterval]` to `def times: List[OgcTimeInterval]`. Now that I think about it, I'm not sure why I didn't do that with the original implementation.

ogc/src/main/scala/geotrellis/server/ogc/OgcSource.scala
```diff
@@ -55,7 +55,7 @@ trait OgcSource {
   def attributes: Map[String, String]
   def resampleMethod: ResampleMethod
   def overviewStrategy: OverviewStrategy
-  def timeInterval: Option[OgcTimeInterval]
+  def timeInterval: OgcTime
```
CloudNiner, line 58: If we do end up keeping `OgcTime`, then this def should probably be renamed as its no longer necessarily an interval. Perhaps `timeInterval` -> `time`?

ogc/src/main/scala/geotrellis/server/ogc/OgcSourceRepository.scala
```diff
@@ -34,15 +34,24 @@ object OgcSourceRepository {
...
     case At(t, _)           => _.filter(_.timeInterval match {
-      case Some(OgcTimeInterval(start, None, _)) => start == t
-      case Some(OgcTimeInterval(start, Some(end), _)) => start <= t && t <= end
+      case OgcTimePositions(list) => list.exists(_ == t)
+      case OgcTimeInterval(start, None, _) => start == t
+      case OgcTimeInterval(start, Some(end), _) => start <= t && t <= end
       // Assume valid OgcLayers with no explicit temporal component are valid
       // at all times so they aren't excluded when optional TIME param is null
-      case None => true
+      case _ => true
```
CloudNiner, line 42: Shouldn't we avoid use of `case _` match everywhere we unpack this ADT? If we use `_` match we wont get exhaustive match errors from the compiler.

ogc/src/main/scala/geotrellis/server/ogc/OgcTime.scala
pomadchin: @CloudNiner WMS also supports TimePoints

ogc/src/main/scala/geotrellis/server/ogc/OgcSourceRepository.scala
pomadchin: `_` stands for `OgcTimeEmpty` here.

pomadchin:
@CloudNiner the current OgcTime tree is the following: 

```scala
sealed trait OgcTime
case object OgcTimeEmpty extends OgcTime
final case class OgcTimePositions(list: NonEmptyList[ZonedDateTime]) extends OgcTime
final case class OgcTimeInterval(start: ZonedDateTime, end: Option[ZonedDateTime], interval: Option[String]) extends OgcTime
```

And you think using `OgcTimeInterval` for everything would be a better idea? 

1. Talking into account that both WMS and WCS services can work with both
2. Query evaluation in case of the OgcTimePositions works as the exact match, while OgcTimeInterval works as a search function on the time range, [see link](https://github.com/geotrellis/geotrellis-server/blob/41a27c3bc7ca2c3ed830f57fde3b390d0218e848/ogc/src/main/scala/geotrellis/server/ogc/OgcSourceRepository.scala#L37-L42)

pomadchin:
I'll work on the interfaces a bit more, but I think that `OgcTimeInterval` should represent an interval where the `OgcTimePositions` should represent a list of times that are can not be reprepsented as an interval. Using the `List[OgcTimeInterval]` sounds like a semantic issue. How would I know how to print such list in the WMS capabilities?

pomadchin:
As an alternative we can have a `newtype` `List[OgcTimeInterval]` which would have a configurable printer.

CloudNiner:
We chatted a bit about this after standup, adding some notes here.

I like the idea of making OgcTimeInterval explicitly an interval and seeing how that falls out, which would make end required in the case class, not sure if `interval` can stay an option or not. Maybe we can avoid the `List[OgcTimeInterval]` case for now and continue to just handle a single OgcTimeInterval? IIRC printing an OgcTimeInterval in the response depends on the spec for each of the response formats.

ogc/src/main/scala/geotrellis/server/ogc/OgcSource.scala
```diff
@@ -55,7 +55,7 @@ trait OgcSource {
   def attributes: Map[String, String]
   def resampleMethod: ResampleMethod
   def overviewStrategy: OverviewStrategy
-  def timeInterval: Option[OgcTimeInterval]
+  def time: OgcTime
```
CloudNiner, line 58: 🎉

ogc/src/main/scala/geotrellis/server/ogc/OgcSource.scala
```diff
@@ -110,36 +131,50 @@ case class GeoTrellisOgcSource(
...
-    else if (source.times.size == 1) OgcTimeInterval(source.times.head).some
-    else OgcTimeInterval(source.times.min, source.times.max.some, None).some
+  lazy val time: OgcTime =
+    if (!source.isTemporal) OgcTimeEmpty
+    else OgcTimePositions(source.times)
 
   /**
    * If temporal, try to match in the following order:
+   *
+   * OgcTimeInterval behavior
```
CloudNiner, line 141: 🎉 updated docstring

ogc/src/main/scala/geotrellis/server/ogc/OgcTime.scala
```diff
@@ -0,0 +1,122 @@
...
+  def apply(times: List[ZonedDateTime]): OgcTime =
+    times match {
+      case head :: tail => OgcTimePositions(NonEmptyList(head, tail))
+      case _ => OgcTimeEmpty
+    }
+  def apply(times: List[String])(implicit d: DummyImplicit): OgcTime = apply(times.map(ZonedDateTime.parse))
+}
+
+/**
+  * Represents the TimeInterval and TimePosition types used in TimeSequence requests
```
CloudNiner, line 72: Docstring should be updated to remove refs to `TimePosition` now that this explicitly represents an interval.

ogc/src/main/scala/geotrellis/server/ogc/OgcTime.scala
```diff
@@ -0,0 +1,122 @@
...
+    else start.toInstant.toString
+}
+
+object OgcTimeInterval {
+  /**
+   * Merge two OgcTimeInterval instances
+   * This semigroup instance destroys the interval. If you need to retain interval when combining
+   *  instances, perform this operation yourself.
+   */
+  implicit val ogcTimePositionsSemigroup: Semigroup[OgcTimeInterval] = Semigroup.instance { (l, r) =>
```
CloudNiner, line 102: implicit name should be updated. Looks like it was copy pasted from OgcTimePositions above

ogc/src/main/scala/geotrellis/server/ogc/params/ParamMap.scala
```diff
@@ -73,27 +74,41 @@ case class ParamMap(params: Map[String, Seq[String]]) {
...
\ No newline at end of file
+
+  def validatedOgcTimeSequence(field: String): ValidatedNel[ParamError, List[OgcTime]] =
+    validatedOptionalParam(field).map {
+      case Some(timeString) if timeString.contains("/") => timeString.split(",").map(OgcTimeInterval.fromString).toList
+      case Some(timeString) => OgcTimePositions(timeString.split(",").toList) :: Nil
+      case None => List.empty[OgcTime]
+    }
+
+  def validatedOgcTime(field: String): ValidatedNel[ParamError, OgcTime] =
```
CloudNiner, line 108: Couldn't this method be simplified to just return the `head` from the result of `validatedOgcTimeSequence`? It would avoid duplication of what looks like the same logic.

ogc/src/main/scala/geotrellis/server/ogc/wms/CapabilitiesView.scala
```diff
@@ -129,20 +129,48 @@ object CapabilitiesView {
...
+                new GridExtent[Long](e, cs)
+              })
+            case rasterOgcLayer: RasterOgcSource =>
+              val rs = rasterOgcLayer.source
+              ReprojectRasterExtent(rs.gridExtent, rs.crs, LatLng, Options.DEFAULT.copy(rasterOgcLayer.resampleMethod))
+          }
+
+          /**
+           * TODO: replace with source.extentIn(LatLng)
+           * see: https://github.com/locationtech/geotrellis/issues/3258
```
CloudNiner, line 148: 😭

stac-example/src/main/scala/geotrellis/stac/raster/StacAssetRasterSource.scala
```diff
@@ -0,0 +1,44 @@
...
+import _root_.io.circe.JsonObject
+import cats.syntax.option._
+
+case class StacAssetRasterSource(
+  asset: StacItemAsset,
+  itemProperties: JsonObject = JsonObject.empty,
+  private[raster] val underlyingRS: Option[RasterSource] = None,
+  private[geotrellis] val targetCellType: Option[TargetCellType] = None
+) extends RasterSource {
+  private lazy val underlying = underlyingRS.getOrElse(RasterSource(asset.href.replace("https://s3-us-west-2.amazonaws.com/landsat-pds/c1/", "gdal+file:///Users/daunnc/Downloads/lc8-test/")))
```
CloudNiner: Don't forget to remove the local ref here.

pomadchin: woooot thanks!

PR #311, opened 2021-02-12
Include default namespace in XML responses

### LINKED ISSUE #298: Missing default XML namespace in WMS/WCS responses
There is no default XML namespace attribute defined on the root element of WMS/WCS GetCapabilities or other XML responses. The default XML namespace attribute is usually used to inform XML parser about the scope of unprefixed XML element in an XML document.

This may cause XML parser to sometime fail to parse properly XML response. For instance, [owslib ](https://github.com/geopython/OWSLib) library cannot parse WMS/WCS GetCapabilies because of this missing attribute.

The expected behaviour would be to explicitly add the default XML namespace attribute on the root element in WMS/WCS GetCapabilities or other XML responses document as suggested below:

## WMS GetCapabilities
`<WMS_Capabilities xmlns="http://www.opengis.net/wms" xmlns:ogc="http://www.opengis.net/ogc" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.3.0">`

## WCS GetCapabilities
`<Capabilities xmlns="http://www.opengis.net/wcs/1.1.1" xmlns:gml="http://www.opengis.net/gml" xmlns:ogc="http://www.opengis.net/ogc" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.1.1">`

 This was successfully tested using static document derived from actual responses.

jisantuc:
## Overview

This PR throws OGC `wcs`, `wms`, and `wmts` namespaces into the `xmlns` attribute for XML responses for each service.

### Checklist

## Notes

Testing could be improved for sure -- I don't know enough about `owslib` right now to do a great job crafting the example calls. Some problems I've run into are not knowing how to make a `GetCapabilities` request to any of the services / not knowing how to translate owslib `Operations` into owslib calls for xml that it wants to parse.

## Testing Instructions

- follow the instructions in `usage.md` to get the `ogc-example` up and running
- however you desire, get a nice isolated python environment, and `pip install` the following requirements file. pyproj is pinned because OWSLib does the `>=` dependency range for pyproj, version 3 of pyproj exists, and version 3 of pyproj depends on a version of libproj past what's available in the default ubuntu ppa.

```
pyproj==2.6.0
OWSLib==0.23.0
```

- fire up a python shell and make some services:

```python
from owslib import wcs, wmts, wms
wcs_service = wcs.wcs111.WebCoverageService_1_1_1("http://localhost:9000", None, None)
wmts_service = wmts.WebMapTileService("http://localhost:9000")
wms_service = wms.WebMapService("http://localhost:9000", version="1.3.0")
```

~~We can't make a `wms` service, because initializing the service with a url throws some kind of error about `'false'` not being a valid int. Unclear where the problem lies there.~~

- make some requests that return xml:
  - `wmts_service.getServiceXML()`
  - `wcs_service.getDescribeCoverage(None)`
  - `wms_service.getServiceXML()`

- `owslib` shouldn't complain about bad XML

Closes #298

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ pc.stdout.log
 .metals
 metals.sbt
 .history
+.bsp
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Fixed
 - Fix STAC API TemporalExtent JSON representation [#293](https://github.com/geotrellis/geotrellis-server/pull/293)
+- XML responses in OGC services include a default namespace [#311](https://github.com/geotrellis/geotrellis-server/pull/311)
 
 ## [4.2.0] - 2020-06-23
 
--- a/ogc-example/docs/usage.md
+++ b/ogc-example/docs/usage.md
@@ -23,9 +23,9 @@ Start SBT
 > ./sbt
 ```
 
-Navigate to the `ogcExample` project
+Navigate to the `ogc-example` project
 ```sh
-root > project ogcExample
+root > project ogc-example
 ```
 
 Fire up the server with some sane defaults for local clients (using 127.0.0.1
--- a/ogc/src/main/scala/geotrellis/server/ogc/wcs/CapabilitiesView.scala
+++ b/ogc/src/main/scala/geotrellis/server/ogc/wcs/CapabilitiesView.scala
@@ -218,7 +218,7 @@ class CapabilitiesView[F[_]: Functor](wcsModel: WcsModel[F], serviceUrl: URL, ex
           ),
           namespace = None,
           elementLabel = "Capabilities".some,
-          scope = constrainedWCSScope,
+          scope = wcsScope,
           typeAttribute = false
         )
         .asInstanceOf[Elem]
--- a/ogc/src/main/scala/geotrellis/server/ogc/wcs/CoverageView.scala
+++ b/ogc/src/main/scala/geotrellis/server/ogc/wcs/CoverageView.scala
@@ -48,7 +48,7 @@ class CoverageView[F[_]: Functor](wcsModel: WcsModel[F], serviceUrl: URL, identi
           obj = CoverageDescriptions(coverageType.values.toList),
           namespace = None,
           elementLabel = "CoverageDescriptions".some,
-          scope = constrainedWCSScope,
+          scope = wcsScope,
           typeAttribute = false
         )
         .asInstanceOf[Elem]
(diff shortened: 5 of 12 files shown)


pomadchin (changes requested):
Thanks, @jisantuc! `WCS` and `WMTS` endpoints now perfectly work.

The story with WMS here is that it generates and incorrect XML. Check out the schema: https://github.com/geotrellis/geotrellis-server/blob/develop/opengis/src/main/xsd/wms/1.3.0/capabilities_1_3_0.xsd#L339

```xml
<attribute name="queryable" type="boolean" default="0"/>
```

Even though it is "boolean" but it encodes boolean as 1 / 0 and not as true / false.

I found this issue: https://github.com/geopython/OWSLib/issues/64

TLDR; that's how the `XMLFormat[Boolean]`  should be implemented:
```scala
/**
   * WMS Service requires a custom boolean writer.
   * Even though WMS works with Booleans:
   *  false should be encoded into the "0" string
   *  true should be encoded into the "1" string
   * */
  implicit lazy val BooleanXMLFormat: XMLFormat[Boolean] = new XMLFormat[Boolean] {
    def reads(seq: scala.xml.NodeSeq, stack: List[ElemName]): Either[String, Boolean] =
      seq.text match {
        case "1" | "true" => Right(true)
        case "0" | "false" => Right(false)
        case x => Left(s"Invalid boolean: $x")
      }

    // check out that we encode boolean as an integer not as a obj.toString
    def writes(obj: Boolean, namespace: Option[String], elementLabel: Option[String],
               scope: scala.xml.NamespaceBinding, typeAttribute: Boolean): scala.xml.NodeSeq =
      Helper.stringToXML(if (obj) "1" else "0", namespace, elementLabel, scope)
  }
  ```
  
However, the approach with defining a custom codec may not work: there is chance that there will be some implicits priorities conflict. In this case my immediate idea would be to fix xsd files here: https://github.com/geotrellis/geotrellis-server/blob/develop/opengis/src/main/xsd/wms/1.3.0/capabilities_1_3_0.xsd#L339, and change all `booleans` to `strings`, i.e.:
  
  ```xml
<!-- <attribute name="multipleValues" type="boolean"/> -->
<attribute name="multipleValues" type="string"/>
<!-- <attribute name="nearestValue" type="boolean"/> -->
<attribute name="nearestValue" type="string"/>
<!-- <attribute name="current" type="boolean"/> -->
<attribute name="current" type="string"/>
```

This would require a minor code change here to make the compiler happy: https://github.com/geotrellis/geotrellis-server/blob/develop/ogc/src/main/scala/geotrellis/server/ogc/wms/CapabilitiesView.scala#L236:

```scala
// replace 
"@queryable" -> DataRecord(false)
// with
"@queryable" -> DataRecord("0")
```

And after that the following test should work (if nothing else is generated wrong):

```python
from owslib import wms
wms_service = wms.WebMapService('http://localhost:9000', version='1.3.0')
wms_service.getServiceXML()
```

pomadchin (approved):
LGTM! Everything works now!

PR #339, opened 2021-03-09
STAC collections support

### LINKED ISSUE #338: STAC Collections support
At this point GeoTrellis server uses the layer extensions to fetch mosaics. What if the target collection has no layers and every collection can be considered as a colelction?

There should be a backwards compatible soft move to support collections istead of layers.

pomadchin:
## Overview

This PR adds backwards compatible collections support in case the STAC Catalog does not use the STAC Layer extension.

### Demo

```javascript
    stac-hillshade-buffered = {
        type = "stacsourceconf"
        // ...
        // this can be collection
        collection = "hillshade-buffered"
        // or layer
        layer = "hillshade-buffered"
        // ...
    }
```

### Checklist

Closes #338

### DIFF
--- a/stac-example/src/main/scala/geotrellis/server/ogc/stac/SearchFiltersQuery.scala
+++ b/stac-example/src/main/scala/geotrellis/server/ogc/stac/SearchFiltersQuery.scala
@@ -73,7 +73,7 @@ object SearchFiltersQuery {
         datetime = left.datetime |+| right.datetime,
         intersects = left.intersects |+| right.intersects,
         collections = (left.collections |+| right.collections).distinct,
-        items = (left.collections |+| right.collections).distinct,
+        items = (left.items |+| right.items).distinct,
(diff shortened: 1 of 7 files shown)


stac-example/src/main/scala/geotrellis/server/ogc/stac/SearchFiltersQuery.scala
```diff
@@ -113,7 +113,7 @@ object SearchFiltersQuery {
         datetime = left.datetime |+| right.datetime,
         intersects = left.intersects |+| right.intersects,
         collections = (left.collections |+| right.collections).distinct,
-        items = (left.collections |+| right.collections).distinct,
+        items = (left.items |+| right.items).distinct,
```
pomadchin, line 116: That is a cruel typo!

pomadchin:
Going to merge it to unblock follow up issues.

Issue #344, opened 2021-03-10
MosaicRasterSource improvements


pomadchin:
Right now, the [MosaicRasterSource](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/MosaicRasterSource.scala#L41) doesn't fetch data from `RasterSources` asynchornosly. This can cause some serious performance issues in case the mosaic is built of 5 or more `RasterSources`.

This issues live in the GeoTrellis server repository since it can be a nice playgroud to implement and test the MosaicRasterSource improvement and after that to backport it into the GeoTrellis core.

PR #362, opened 2021-04-23
MosaicRasterSource improvements

pomadchin:
## Overview

This PR adds effectful RasterSources concept into the repo as well as two MosaicRasterSources: one that works only with effectful RasterSources, another one is able to lift RasterSource into the context.

This PR resolves https://github.com/geotrellis/geotrellis-server/issues/344 but does not address the general issue that RasterSources as synchrnous interfaces. It contains some attempts and backports from the geotrellis-contrib repo, but it became obvious that it would require much more time to accomplish them.

### Checklist

Closes https://github.com/geotrellis/geotrellis-server/issues/344


effects/src/main/scala/geotrellis/raster/effects/MosaicRasterSourceF.scala
```diff
@@ -0,0 +1,235 @@
...
+/** Single threaded instance of a reader for reading windows out of collections
+  * of rasters
+  *
+  * @param sources The underlying [[RasterSourceF]]s that you'll use for data access
+  * @param crs The crs to reproject all [[RasterSourceF]]s to anytime we need information about their data
+  * Since MosaicRasterSources represent collections of [[RasterSourceF]]s, we don't know in advance
+  * whether they'll have the same CRS. crs allows specifying the CRS on read instead of
+  * having to make sure at compile time that you're threading CRSes through everywhere correctly.
+  */
+abstract class MosaicRasterSourceF[F[_]: Monad: Parallel] extends RasterSourceF[F] {
```
pomadchin, line 44: Sadly, none of these interfaces were integrated into gt server :/

PR #395, opened 2023-11-10
WMTSModel add GeoTrellisOgcSource match item

RunBoo:
## Overview

WMTS does not support GT Layer now, it's necessary to add GeoTrellisOgcSource to WMTSModel.


pomadchin (approved):
:+1:

